PR #18, opened 2026-07-16
Add native CPU/CUDA LoRA training and restart-safe resume

betweentwomidnights:
## Summary

This integrates native Stable Audio 3 LoRA/DoRA training into the C++ CLI.

- adds the `sa3-train` workflow for medium-base, small-music-base, and small-sfx-base
- matches the verified Stable Audio 3 training/noising/conditioning behavior
- adds exact restart-safe resume state
- adds CPU and CUDA training support, with Vulkan in active development and Metal planned afterward
- adds training-base model conversion, downloads, model cards, and training documentation
- pins the reviewed `betweentwomidnights/ggml` patch stack needed for training autodiff
- preserves the existing inference CLI, server, and embedded-library workflows
- documents and detects stale submodule checkouts for existing clones and downstream forks

## Attribution and history

This integration supersedes and incorporates #10. @Bakobiibizo's original authored commits remain in the branch history so his contribution and authorship are preserved.

Please merge this PR with a merge commit or fast-forward; do not squash it.

Closes #10.

## Backend status

- CPU training: tested
- CUDA training: tested
- Vulkan inference: retained; training shader work is next
- Metal training: planned after Vulkan

Inference behavior should remain unchanged. The trainer is intentionally documented as active development while backend coverage expands.

## Validation

- Windows CPU Release build
- Windows CUDA Release build
- CPU CTest: 16/16 passed
- CUDA CTest: 16/16 passed
- medium-base and small-base native training runs
- inference application of generated adapters
- uninterrupted versus resumed CUDA training produced identical adapter and state hashes
- CUDA checkpoint successfully resumed on CPU
- current branch is based on the latest `main`
- existing-fork migration reproduced against `pillopaus-project/sa3.cpp` at both `2742515` and `a50048a`

## Downstream-fork check requested

The one-time URL migration must fetch the superproject without recursive submodule fetching, then synchronize the cached URL:

```sh
git -c fetch.recurseSubmodules=false fetch upstream
git merge upstream/lora-training-integration
git submodule sync --recursive
git submodule update --init --recursive
```

Fresh clones are expected to work directly with `--recurse-submodules`. CMake warns when the local ggml checkout does not match the exact pin without modifying local ggml work.


betweentwomidnights:
@pillopaus-project, when you have a moment, would you be willing to test this draft from an existing checkout of your web UI fork?

The most useful downstream compatibility check would be:

1. fetch this `lora-training-integration` branch into a temporary test branch in your fork;
2. run:
   ```sh
   git submodule sync --recursive
   git submodule update --init --recursive
   ```
3. rebuild and launch the inference/server path your web UI currently uses;
4. let us know whether the ggml URL/pin transition or any existing CLI/server behavior causes trouble.

Fresh clones are also expected to work with `--recurse-submodules`, but the existing-checkout path is the one I especially want to validate before merging. No need to exercise LoRA training itself unless you feel adventurous... the key thing is that your current downstream integration still syncs, builds, and runs normally. Thank you!

pillopaus-project:
@betweentwomidnights 
my fork at this moment in time is diverging quite a bit from your repo.
the ENV_VARS use imho is problematic, as stale env_vars can stay active and cause problems ( happened a few times already while testing different git repos with different directories. thats why i changed  as follows :

**Replaced all "SA3_*" env vars with Sa3Config CLI flags + JSON params**
- Created src/sa3_config.h as single source of truth for all defaults
- Removed all getenv("SA3_*") calls; all env fallback logic eliminated
- Removed cpu_threads_from_env(), loudness_defaults_from_env(), profile_enabled()
**New CLI flags** ( valid for both sa3-server, sa3-generate):
  --device, --gpu, --flash-attn, --same-flash-attn, --profile, --dump-cond
  --models-dir optional, defaults to ./models
**Runtime JSON params** (for server : /generate, /generate/loop):
  loop_pad_seconds, default_loop_bars, max_duration, bpm, seed,
  inpaint_start, inpaint_end, all CFG fields, all loudness fields,
  encode/decode chunk size + overlap

As im still working on it  the best thing would be for you to clone my repo and pin it to the a50048a54b03b40b029d634fef974b5c00acaf2a  push (before the env change), then to merge your branch locally and check it

have a look and let me know
PP

betweentwomidnights:
Thanks PP 

I tested the integration locally against both a50048a and its parent.
One clarification: a50048a appears to be the environment-variable rewrite itself. Its parent, 2742515, looks like the actual pre-env-change baseline, so I tested both.
The submodule transition itself works, but we found an important migration detail: the initial fetch must disable recursive submodule fetching because an existing clone still has the old ggml-org/ggml URL cached:

```
git -c fetch.recurseSubmodules=false fetch upstream
git merge upstream/lora-training-integration
git submodule sync --recursive
git submodule update --init --recursive
```

Your fork also has branch = master in .gitmodules, so the merge produces one small expected conflict there. I resolved it by selecting https://github.com/betweentwomidnights/ggml.git and removing the moving branch setting. After that, ggml synced correctly to the expected cfec69c pin.

On the pre-env baseline, all 16 tests passed and the generate/training CLIs built. The Windows build only failed on sa3-server with MSVC C2026 errors from the large string literals in src/embedded_web.h. The training integration obviously doesn't touch that file, and the same server error occurs when testing the a50048a baseline.

Anyway, the env/config refactor sounds interesting and i'd probably be down to incorporate that on main once I finish with the LoRA training integration. 

No big deal if you're planning on keeping your fork independent and want to continue using the official ggml instead. The forked ggml pin is only relevant if you want to pull in the LoRA training stuff or main down the road.

pillopaus-project:
sorry for the wrong push ref , i copied the wrong one...
regarding the ggml version, next im going to see if acestep ggml fork can be useful for some speed gain... will report later on...
PP

betweentwomidnights:
yeah i initially didn't want to branch off of standard ggml but now that i have i'm diving into shaders and stuff to optimize the training as much as possible on iGPUs. Hoping some of these optimizations will apply to inference time as well.

PR #21, opened 2026-07-17
train: validate native Metal LoRA training

betweentwomidnights:
## What changed

- Pin the immutable ggml `sa3-training-v1-metal` milestone after native backward-op support and `OUT_PROD` optimization.
- Preserve the original validated DiT inference graph and use a separate autodiff-safe graph for training, fixing the temporary corrupted base-audio regression without changing the training trajectory.
- Add macOS training portability fixes and opt-in checkpoint-segment profiling.
- Record Metal correctness, MLX A/B, and the completed 2,000-update Ratatat benchmark.

## Root cause of the inference regression

The first training implementation rewrote shared DiT operations for autodiff safety. Those explicit repeats, contiguous operands, expanded sigmoid, and memory-token construction also changed the inference graph's fusion and numerical behavior. Training now opts into those forms explicitly; normal generation retains the original graph.

## Measured result on Apple M4

- Scalar Metal baseline: 22.364 s/update.
- Optimized short-run steady sample: 8.649 s/update.
- Full 2,000-update sustained average (steps 2-2,000): 9.895 s/update.
- Full wall time including startup and four checkpoint pairs: 5 h 29 min 55 s.
- Maximum RSS: 5.76 GiB; peak footprint: 5.52 GiB.
- Matched MLX short-run reference: 7.285 s/update and 16.27 GiB maximum RSS.

## Validation completed

- 37/37 complete tests passed.
- Metal `OUT_PROD`: 92/92 focused cases.
- Final adapter is byte-identical to the step-2,000 adapter.
- All 2,000 loss and gradient-norm records are finite.
- The retained medium base WAV is byte-identical after the inference/training graph split.
- User ear checks confirmed healthy base inference and recognizable Ratatat adaptation across multiple in-distribution prompts.

## Windows regression audit

- CPU, CUDA, and Vulkan builds completed successfully; all 16 registered tests passed on each backend.
- Base medium generation passed across all three backends (raw waveform cosine 0.991-0.995).
- Medium LoRA generation passed across all three backends (raw waveform cosine 0.997-0.999).
- Matched three-step CUDA and Vulkan training replays reproduced every loss and gradient norm and wrote byte-identical adapter and trainer-state GGUFs versus their pre-split references.
- Vulkan `OUT_PROD` passed 91/91 cases on both the NVIDIA dGPU and Intel iGPU.

## Fresh-clone packaging gate

A clean recursive clone checked out parent `d9044b5` and immutable ggml pin `922875a6` from the public fork, with the expected submodule URL and no moving branch setting. A cold Windows CPU build completed for the generation CLI, training CLI, server, shared library, and test executables; all 16 registered tests passed. Both worktrees remained clean.

No validation blocker remains.


betweentwomidnights:
Windows post-Metal audit at parent `0f05f0f` / ggml `922875a6`:

- CPU, CUDA, and Vulkan builds completed successfully; all 16 registered tests passed on each backend.
- Base medium generation passed on all three backends. Raw waveform cosine: CPU/CUDA `0.995041`, CPU/Vulkan `0.994372`, CUDA/Vulkan `0.991467`.
- Medium LoRA generation passed on all three backends. Raw waveform cosine: CPU/CUDA `0.999094`, CPU/Vulkan `0.997482`, CUDA/Vulkan `0.996710`.
- A matched 3-step CUDA medium replay reproduced every loss and gradient norm and produced byte-identical adapter and trainer-state GGUFs versus the pre-inference-split reference.
- A matched 3-step Vulkan small-music replay did the same; both serialized outputs were byte-identical to their references.
- Expanded Vulkan `OUT_PROD` coverage passed 91/91 on both the NVIDIA dGPU and Intel iGPU.
- Call-site audit confirms inference callers retain the default `autodiff_safe=false`, while every training graph builder explicitly passes `true`.
- The temporary `CODEX_METAL_TRAINING_HANDOFF.md` has been removed, and stale README/training-guide wording has been updated.

No Windows regression or merge-blocking finding surfaced in this pass.

PR #23, opened 2026-07-25
quant: Q4_K_M / Q8_0 quantization tooling

betweentwomidnights:
Ports the Q4_K_M quantization tooling from @pillopaus-project's
[`q4km-webapp-fixes`](https://github.com/pillopaus-project/sa3.cpp/tree/q4km-webapp-fixes)
branch, stripped of the web app so `main` stays primitives-only. Scoped deliberately
tightly — his web interface, config/env rework and trainer UI are separate later stages —
so he can point his `.gitmodules` back at our ggml fork.

**Depends on betweentwomidnights/ggml#2** (pinned here as submodule `e9c70fd6`).

## What lands

- `sa3-quantize`, `sa3-quant-check`, `sa3-quant-eval`, and their CMake targets.
- `--mix q4_k_m | q5_k_m | q5_k | q8_0` — each names the type used for ordinary 2D weights
  and the type for the V/down/embed tensors that get promoted.
- `ModelPaths::resolve` and `sa3-generate` accept `q4_k_m` / `q5_k_m` / `q5_k` / `q8_0`.

The resolver resolves the requested encoding **exactly** rather than falling through a
candidate list. The upstream port fell back silently, so `--encoding f16` with only Q4_K_M
present would load Q4_K_M and say nothing. This repo exists for numeric parity; quietly
substituting a precision is worse than an error. A miss now names which encodings do exist
for that variant.

## Bugs fixed on top of the port

1. **`sa3-generate` ignored `--encoding` entirely.** It kept a private copy of the
   suffix mapping, `(strcmp(encoding,"f32")==0) ? "F32" : "F16"`, collapsing every other
   value to F16 — so `--encoding q4_km` generated from the F16 weights and produced
   *bit-identical* output. Pre-existing on `main`, and present on pillo's branch too; he
   only ever exercised Q4 through the server and `sa3-quant-eval`, which both go through
   `ModelPaths::resolve`. Now shares `sa3::encoding_suffix()` and rejects unknown values.
2. **`file_size()` used `ftell`, whose `long` is 32-bit on MSVC.** Any gguf ≥ 2 GiB got a
   bogus length and crashed before the first tensor (`0xC0000409`). The medium DiT is
   2.7 GiB at F16, so medium could not be quantized at all on Windows; every sub-2 GiB file
   worked, which hid the boundary. Uses `std::filesystem::file_size`, reads in chunks, and
   reports a short read with its offset.
3. **`should_quantize()` gated rows on `ne[0] % 32`,** but Q4_K/Q5_K/Q6_K are k-quants with
   256-element super-blocks. `ne[0]` of 32/64/128 passed and then overran the destination
   rows. A rank-32 LoRA (`ne[0] == 32`) is the case that shows up in practice. Now gated on
   the target type's own block size.
4. **Designated initializers** require C++20; this tree is C++17.

Naming follows the Encoding field already fixed in `docs/DISTRIBUTION.md` (`Q4_K_M`, not
`Q4_KM`), matching the gguf spec and llama.cpp. Compact spellings stay accepted as input
aliases; only the canonical form is written or resolved.

## Results

Sizes (DiT + SAME):

| model | F16 | Q8_0 | Q5_K_M | Q4_K_M |
| --- | --- | --- | --- | --- |
| small-music | 1083 MB | 591 | 422 | 384 |
| medium | 4399 MB | 2348 | 1671 | 1532 |

Speed, 30 s / 8 steps, best of 3 warm:

| backend | model | F16 | Q8_0 | Q5_K_M | Q4_K_M |
| --- | --- | --- | --- | --- | --- |
| CUDA | small-music | 2.36 | 2.03 | 1.98 | 1.98 |
| Vulkan | small-music | 3.08 | 2.47 | 2.35 | 2.32 |
| CPU | small-music | 30.77 | 20.86 | 22.27 | 19.32 |
| CUDA | medium | 5.54 | 4.15 | 3.71 | 3.65 |
| Vulkan | medium | 6.61 | 5.23 | 4.62 | 4.50 |
| CPU | medium | 122.70 | 79.68 | 84.32 | 72.48 |
| Metal (M4) | medium | 14.51 | 14.27 | — | 14.81 |

Q4_K_M is ~33% faster than F16 on both discrete-GPU backends and 41% faster on CPU for
medium. **Metal is the exception: flat.** Q8_0 is 1.7% faster there and Q4_K_M 2.1%
*slower* — unified memory means there is no host-to-VRAM copy to shrink (`dit_upload` is
0.5 ms), so the 0.9 s of load-time saving is cancelled by k-quant dequant work against
Metal's F16 simdgroup-matrix `mul_mm` path (`dit_compute` +10.4%). On Apple silicon this
is a footprint optimization only. Full breakdown in `docs/METAL.md` section 9. `sa3-quant-check`
reports `below-threshold=0` at ≥0.99 cosine for all nine model × encoding combinations.

Q4_K_M and Q8_0 are the two worth publishing. `q5_k_m` is speed-neutral against `q4_k_m` on
GPU but 9% larger, and on CPU it loses to `q8_0` on *both* speed and fidelity, so it has no
clear niche. `q5_k` is a coverage-only mix (see below), never to be published.

## Validation

All four backends now covered — CUDA, Vulkan, CPU (Windows) and Metal (M4, `docs/METAL.md`
section 9). Metal specifics:

- Quantized outputs are **byte-identical** to the Windows ones, confirming quantization is
  host-side and reproducible across hosts.
- `sa3-quant-check` `below-threshold=0` for DiT (204) and SAME-L (100), Q8_0 and Q4_K_M.
- The ggml pin is confirmed **inert on Metal**: an F16 render on this branch is byte-identical
  to one built from `main` (`md5 cfd0e775d12a6115661768e717db00a0`) at the same warm speed.
- F16 vs Q4_K_M waveform cosine 0.529, rms envelope 0.956 — matching the ~0.55 seen
  elsewhere. Diffusion trajectory shift, judged by ear rather than by cosine.

The `q5_K` `get_rows` fix in ggml#2 now has four independent confirmations: the corrected CUDA
kernel agrees with the canonical CPU path (0.987014 / 0.986894), with Vulkan's shader
(0.988893) and with Metal's `kernel_get_rows_q5_K` (0.988871).

Unrelated finding logged while measuring: Metal F16 at this shape is 14.5 s against 12.14 s
recorded in `docs/METAL.md` section 6 (2026-06-29). The gap reproduces on `main`, so it
predates this work — it arrived with the ggml v0.16.0 base move plus the Metal training merge.
Flagged for a separate bisect, not attributable to quantization.

## Note on the ggml dependency

The submodule bump is **not** required for anything in this PR's release shortlist:
`te.embed.weight` (`src/t5gemma.h:47`) is the only `ggml_get_rows` input in the pipeline and
it lives in the T5 encoder gguf, so quantized DiT/SAME never reach it. The pin matters only
for a quantized encoder in a k-quant mix. A Q8_0 encoder needs no patch at all (Q8_0 was
already in CUDA's dispatch) and is near-lossless at 0.9993 cosine vs F32 — likely the better
encoder option anyway, at 1074 MB → 285 MB.


betweentwomidnights:
## M4 / Metal validation — all gates pass

Ran the `METAL_QUANT_HANDOFF.md` gates on the M4 (32 GB, macOS 15). Results are in
`docs/METAL.md` section 9 (c3a54c2); the handoff doc is deleted, as it asked to be.
**Validated here — ready to review and merge from the PC.**

Headline: **on Metal, quantization is a size win but not a speed win.**

| backend | F16 | Q8_0 | Q4_K_M |
|---|---:|---:|---:|
| CUDA (RTX 5070 laptop) | 5.54 | 4.15 | 3.65 |
| Vulkan (same GPU) | 6.61 | 5.23 | 4.50 |
| **Metal (M4)** | **14.51** | **14.27** | **14.81** |

Q8_0 is 1.7% faster than F16 and Q4_K_M is 2.1% *slower*, against ~33% gains on the other two
GPU backends. Run-to-run spread was under 0.2 s, so this is signal rather than noise.

`SA3_PROFILE=1` explains it. Quantization does save load time — 0.63 s at Q8_0, 0.91 s at
Q4_K_M — but it costs `dit_compute` +3.8% / +10.4% and `dec_compute` +2.6% / +7.0%. The two
effects cancel. Unified memory is the reason: `dit_upload` is **0.5 ms** because there is no
copy to shrink, and at 324 frames the DiT runs the ALU-bound `mul_mm` path, where dequantizing
k-quant blocks is pure added work against Metal's F16 simdgroup-matrix kernels.

This doesn't change the merge decision — 4399 → 1532 MiB is still the reason to ship it on
Apple silicon — but we shouldn't advertise a speedup on Mac.

### Gates

- **1, unregressed.** F16 output is byte-identical to a build from `main`
  (`md5 cfd0e775d12a6115661768e717db00a0`), at the same warm speed (14.51 vs 14.53 s). The ggml
  pin is confirmed inert on Metal both by diff — it touches only `src/ggml-cuda/*` plus two
  additive `gguf_get_tensor_*` accessors — and by measurement.
- **2, sizes.** 962.43 / 569.90 / 1483.36 / 864.65 MiB, matching the Windows table exactly.
  Quantization is reproducible across hosts.
- **3, quant-check.** DiT `compared=204 below-threshold=0`, SAME-L `compared=100
  below-threshold=0`, at `threshold=0.990`, for both encodings.
- **4, ear test.** All three renders are music of the same character. F16 vs Q4_K_M waveform
  cosine 0.529 with the envelope holding at 0.956; Q8_0 near-transparent at 0.990 / 0.9996.
  The three outputs have three distinct hashes, which incidentally confirms the `--encoding`
  fix — before it these were bit-identical.
- **5, speed.** Table above.

### The q5_K fix, validated from a fourth angle

Encoder hidden-state cosine vs F32, `--seed 7`:

| encoder embed | CUDA | CPU | Vulkan | **Metal (M4)** |
|---|---:|---:|---:|---:|
| Q8_0 | 0.999301 | 0.999242 | 0.999641 | **0.999686** |
| Q6_K (`q4_k_m`) | 0.979378 | 0.978046 | 0.980380 | **0.978471** |
| Q5_K (`q5_k`) | 0.987014 | 0.986894 | 0.988893 | **0.988871** |

Metal's `kernel_get_rows_q5_K` is an independent implementation from the CUDA one corrected in
`e9c70fd6`, and it agrees to 0.002 — effectively on top of Vulkan. Had the index mapping still
been wrong, CUDA would read as the outlier. An end-to-end `sa3-generate` with a Q5_K-embed
encoder ran clean.

### One unrelated finding

F16 at this shape is 14.5 s against the 12.14 s recorded in `docs/METAL.md` section 6
(2026-06-29). The gap reproduces on `main`, so it predates this PR — most likely the ggml base
move to v0.16.0 plus the Metal training merge (#19 / #21). Tracking separately; not a blocker
here.

### Before merge

- **Merge, don't squash** on betweentwomidnights/ggml#2 — @pillopaus-project's authorship on
  `74e73f67` needs to survive, and `e9c70fd6` must stay reachable as the submodule pin.
- Minor cosmetic: `sa3-quantize --mix q8_0` prints `(204 q8_0, 0 q8_0, …)`, reusing the base
  type name in the promoted-type slot. Harmless.

PR #24, opened 2026-07-25
functional LoRA: adapters on a quantized base, composed across a chain

betweentwomidnights:
Ready to merge. Every item on the original checklist is closed: Metal measured, multi-adapter
composition implemented and tested, the `-xs` families added, and the branch rebased onto
`15dd87a`. Final validation was re-run on the PC (CUDA, RTX 5070) after the rebase.

## What this does

A LoRA could not be applied to a quantized base at all. `apply_loras` rewrites `W` in place,
which means reading `W` out and writing `W_eff` back, and neither direction has a route for
k-quant types — the graph path aborts in `ggml_cast` (`unsupported type combination
(q6_K to f32)`) and the host path cannot re-pack an f32 result into Q4_K storage. So
`--lora X --encoding q4_k_m` was a hard abort.

This hands the adapter tensors to `dit_lin` instead, which folds them into the graph as
`y = W@x + s·B@(A@x)` plus the dora row scale. `W` is then only ever an argument to `mul_mat`,
and quantized `mul_mat` works on every backend, so the base is never dequantized or rewritten.
`dit_lin` already did exactly this for the training graph; inference just passed `dl = nullptr`.

## Multiple adapters compose exactly

Blending several adapters is central to how these models actually get used, so the functional
path composes a whole chain rather than handling one adapter. Merging sequentially — each
adapter reading the previous one's `W_eff`, which is what `apply_loras` does — expands with **no
approximation** into

```
y = c_0 ⊙ (W@x) + Σ_k s_k · c_k ⊙ (B_k @ (A_k @ x))
c_0 = Π_{dora j} d_j        c_k = Π_{dora j, j >= k} d_j
```

with `d_j` the dora row scale of adapter `j`. Every `c` is constant at inference, so they are
precomputed once and folded into the term tensors: the per-step graph costs one broadcast
multiply plus the residuals, regardless of how many adapters stack. A later dora rescaling
earlier residuals is exactly why the chain does not commute, and the suffix scan preserves that.

"dora does not commute" was the original reason for the one-adapter limit. That was true but it
never implied an approximation was needed — non-commuting only means order matters, and the
recursion keeps the order.

`tests/lora_compose_test.cpp` pins the composed form against the **real merge path** over 16
chains, so the test references `apply_loras` rather than my own algebra.

The precompute reuses what one adapter already needed: `base_norm_sq` (unchanged, still
adapter-independent, so the sidecar stays valid), one `mul_mat(W, A_k)` per adapter, and one
rank×rank gram per adapter *pair*. A purely additive chain needs no norms at all and skips the
host read entirely.

This also covers the `-xs` families, since `U @ M_xs @ Vᵗ` is still rank-r and regroups into the
same terms.

## Results

medium, 20 s output, 8 steps, kev (dora-rows, rank 16), warm, base-norm cache primed:

| backend | path | dit_compute | total | peak VRAM |
| --- | --- | --- | --- | --- |
| CUDA | f16 + merged | 699 ms | 5653 ms | 2994 MiB |
| CUDA | q4_k_m + functional | 728 ms | **3545 ms** | **1278 MiB** |
| Vulkan | f16 + merged | 523 ms | 6674 ms | 3049 MiB |
| Vulkan | q4_k_m + functional | 712 ms | **4310 ms** | **1248 MiB** |
| Metal M4 | f16 + merged | 4925 ms | 13520 ms | 3071 MiB |
| Metal M4 | q4_k_m + functional | 6340 ms | **11720 ms** | **1973 MiB** |

37% faster end to end on CUDA, 35% on Vulkan, 13% on Metal, and ~58%/~59%/36% less memory.

**The win is not per-step.** `dit_compute` is *worse* functionally — +4% CUDA, +36% Vulkan,
+22% Metal — because the residual matmuls are real per-step work the merged path does once up
front. The end-to-end gain comes entirely from loading a 1532 MiB model instead of a 4399 MiB
one.

Two composed adapters (kev + keygen, both dora-rows rank 16, 228 targets, CUDA, 20 s / 8 steps):

| path | dit_compute | total | peak VRAM |
| --- | --- | --- | --- |
| f16 + merged | 816 ms | 6724 ms | 3213 MiB |
| q4_k_m + functional | 1033 ms | **4230 ms** | **1423 MiB** |

37% faster end to end and 56% less VRAM, with a bigger per-step cost (+27%) than one adapter
because there are now two residuals per target.

**On Metal at 8 steps it already inverts, even on f16**: merging costs ~2.4 s up front there,
and at 8 steps that outweighs the accumulated residual cost.

| steps (Metal M4, f16) | merged | functional |
|---:|---:|---:|
| 8 | 13.52 s | **12.18 s** |
| 32 | 29.30 s | 29.97 s |

Crossover is ~20–25 steps and 8 is the common case, but the default is deliberately left alone:
one machine and one adapter type is not enough to move it, and `dora-rows` merging also pays for
dora row work a plain `lora` merge would not. If the crossover holds on CUDA/Vulkan too, the
heuristic wants to be step-count aware rather than dtype-aware, which is its own change.

## Accuracy

Judge this **per forward pass, not per generation**. At one diffusion step, functional vs merged
on the same f16 base is **0.999878 (one adapter) and 0.999876 (two adapters)** — composition adds
no error.

Over 8 steps the same comparison lands anywhere from 0.977 to 0.9995, because a ~1e-4 per-pass
difference compounds into a different-but-valid trajectory. Those numbers move with prompt and
seed and should not be read as a quality ordering: Metal measured 0.999711, CUDA 0.991400,
Vulkan 0.950966, against 0.9968 for identical weights across CUDA/CPU and 0.9714 for medium f16
across CUDA/Vulkan. Vulkan is simply looser between any two paths.

## Two things that made it viable

- **Precomputed row scale** (`d1e2a69`). `dit_lin` recomputed the dora-rows norms every step via
  `mul_mat(W, A)` over the full base weight — necessary while training, where A and B move, but
  at inference `W`, `A` and `B` are all frozen, so the row scale is a constant. Computing it once
  dropped `dit_compute` from 885 ms to 722 ms (+47% → +20%). The one-shot graph reuses
  `dit_lin`'s exact op sequence so the shortcut cannot drift from the reference. Training leaves
  it null and is untouched.
- **Cached base norms** (`e2bec09`). `base_norm_sq` needs every targeted weight pulled to the
  host and squared, ~1.5B elements for the medium DiT: +3.4 s on q4_k_m, +4.6 s on f16 (worse
  there — larger weights to transfer, so not a quantization problem). It depends only on the base
  weights, never the adapter, so it is cached in `<model>.gguf.norms` keyed on the model's byte
  size. First run 6647 ms, every run after 3512 ms. 3.3 MB sidecar for 228 tensors, and
  backend-independent — a cache written by CUDA is reused by Vulkan. Written temp-then-rename so
  an interrupted run cannot leave a half-written cache that later reads as valid.

## No behaviour change to existing paths

- f16/f32 bases still **merge by default**; the merge costs nothing per step. The functional path
  engages only when the base is quantized, or when `SA3_FUNCTIONAL_LORA` is set, to A/B the two
  on f16.
- Supported functionally: the row-normalizing families — `lora`, `dora-rows` and their `-xs`
  variants — in any number and any mix. `dora-cols` and `bora` still merge, so they need f16/f32.
  They normalize over `in`, so their scale lands on the *input*, needing a column norm of the base
  weighted by the row scales accumulated so far — chain-dependent, so unlike `base_norm_sq` it
  cannot be cached per base — plus cross terms contracting against `Wᵗ`, which quantized `mul_mat`
  does not support.
- Training is untouched.
- What was a raw ggml abort on `--lora` + quantized is now either a working generation or a clear
  error naming the constraint.

`--mix f16/f32` on `sa3-quantize` came along for the ride: it re-encodes an already-quantized
gguf, so dequantizing Q4_K_M back to F16 yields a model that is numerically the quantized one but
readable by code that cannot handle quantized tensors. That is how the quality question was
answered before this path existed, and it is what proved the round-trip is a faithful proxy
(0.995400 vs 0.995778 against true F16 on a single forward pass).

## Fixes folded in

- `SA3_FUNCTIONAL_LORA` was a bare `getenv` presence test, so `=0` *enabled* it. Now compares
  against `"0"`.
- `lora.h` referenced `DitLora`/`DitLoraParam` without including `dit.h`, compiling only where an
  earlier include happened to pull it in first. #27 changed that ordering and broke
  `sa3-train-checkpoint-test` with `unknown type name 'DitLora'` on rebase. Fixed in `9bc9533`.

## Validation

Re-run on the PC after the rebase, CUDA / RTX 5070:

- 17/17 ctest pass, including the new `sa3-lora-compose-test` and `sa3-train-checkpoint-test`.
- `--encoding q4_k_m --lora kev` → `lora: applied 1 adapter(s) [functional]`, real audio out.
- `--encoding q4_k_m --lora kev --lora keygen` → `lora: applied 2 adapter(s) [functional]`.
- `--encoding q4_k_m --lora <bora>` → rejected with the constraint message, as intended.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,12 +28,13 @@ bench/
 loras/*
 !loras/README.md
 
-# Model weights / checkpoints
+# Model weights / checkpoints (*.norms = derived base-norm cache, regenerated on demand)
 models/*
 !models/README.md
 checkpoints/
 train-runs/
 *.gguf
+*.norms
 *.safetensors
 metrics.jsonl
 config.snapshot.txt
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -226,6 +226,9 @@ target_link_libraries(sa3-train-inpaint-test PRIVATE sa3)
 add_executable(sa3-dit-lin-functional-test tests/dit_lin_functional_test.cpp)
 target_link_libraries(sa3-dit-lin-functional-test PRIVATE sa3)
 
+add_executable(sa3-lora-compose-test tests/lora_compose_test.cpp)
+target_link_libraries(sa3-lora-compose-test PRIVATE sa3)
+
 if(BUILD_TESTING)
     find_package(Python3 COMPONENTS Interpreter QUIET)
     add_test(NAME sa3-train-config-test COMMAND sa3-train-config-test)
@@ -243,6 +246,7 @@ if(BUILD_TESTING)
     add_test(NAME sa3-train-prompt-test COMMAND sa3-train-prompt-test)
     add_test(NAME sa3-train-inpaint-test COMMAND sa3-train-inpaint-test)
     add_test(NAME sa3-dit-lin-functional-test COMMAND sa3-dit-lin-functional-test)
+    add_test(NAME sa3-lora-compose-test COMMAND sa3-lora-compose-test)
     if(Python3_Interpreter_FOUND)
         add_test(NAME sa3-model-artifacts-test
                  COMMAND "${Python3_EXECUTABLE}" "${CMAKE_CURRENT_SOURCE_DIR}/tests/model_artifacts_test.py")
--- a/docs/BENCHMARKS.md
+++ b/docs/BENCHMARKS.md
@@ -98,6 +98,57 @@ generation is basically a base generation plus ~0.1s per adapter. the in-place p
 a separate copy of the dora'd dit (~2.9gb) would push vram back over 8gb and thrash. on the
 host loops this same apply was ~14s.
 
+### functional apply — the path a quantized base needs (experimental)
+
+the in-place merge cannot run on a quantized base at all. it has to read `W` out and write
+`W_eff` back, and neither direction has a route for k-quant types: the graph path aborts in
+`ggml_cast` (`unsupported type combination (q6_K to f32)`) and the host path cannot re-pack an
+f32 result into Q4_K storage. so `--lora X --encoding q4_k_m` used to be a hard abort.
+
+the alternative is to not merge at all — hand the adapter tensors to `dit_lin` and let it fold
+them into the graph as `y = W@x + s·B@(A@x)` plus the dora row scale. `W` is then only ever an
+argument to `mul_mat`, and quantized `mul_mat` works everywhere. this is the same formulation
+the training graph already used; inference simply passed `dl = nullptr`. see `src/lora.h` for
+the derivation and the limits.
+
+medium, 20s output, 8 steps, kev (dora-rows, rank 16), warm, base-norm cache primed:
+
+| backend | path | dit_compute | total | peak vram |
+|---|---|---|---|---|
+| cuda | f16 + merged | 699 ms | 5653 ms | 2994 MiB |
+| cuda | q4_k_m + functional | 728 ms | **3545 ms** | **1278 MiB** |
+| vulkan | f16 + merged | 523 ms | 6674 ms | 3049 MiB |
+| vulkan | q4_k_m + functional | 712 ms | **4310 ms** | **1248 MiB** |
+
+37% faster end to end on cuda, 35% on vulkan, and ~58% less vram on both. note the win is not
+per-step: `dit_compute` is *worse* functionally (+4% cuda, +36% vulkan), because the residual
+matmuls are real per-step work that the merged path does once up front. the end-to-end gain
+comes from loading a 1532 MiB model instead of a 4399 MiB one. vulkan pays more per step than
+cuda for the same graph, so on a backend where load time did not dominate this could invert.
+
+two things make it viable rather than merely possible:
+
+- **precomputed row scale.** `dit_lin` recomputes the dora-rows norms every step via
+  `mul_mat(W, A)` over the full base weight, which it must do while training because A and B
+  move. at inference `W`, `A` and `B` are all frozen, so `magnitude/||W + s·A@B||_row` is a
+  constant. computing it once cut `dit_compute` from 885 ms to 722 ms (+47% → +20%).
+- **cached base norms.** `base_norm_sq` needs every targeted weight pulled to the host and
+  squared, ~1.5B elements for the medium dit: +3.4 s on q4_k_m and +4.6 s on f16 (worse there,
+  larger weights to transfer — not a quantization problem). it depends only on the base
+  weights, so it is cached in `<model>.gguf.norms`, keyed on the model's byte size. first run
+  6647 ms, every run after 3512 ms. the sidecar is 3.3 MB for 228 tensors and is
+  backend-independent — a cache written by cuda is reused by vulkan.
+
+output is unchanged. functional vs merged on the *same* f16 base is 0.991400 waveform cosine
+on cuda and 0.950966 on vulkan, against 0.9968 for identical weights across cuda/cpu and
+0.9714 for medium f16 across cuda/vulkan. vulkan is simply looser between any two paths; both
+sit far above the ~0.5 floor that two genuinely different trajectories produce.
+
+f16/f32 bases still merge by default — it costs nothing per step. the functional path engages
+when the base is quantized, or when `SA3_FUNCTIONAL_LORA` is set to a/b the two on f16.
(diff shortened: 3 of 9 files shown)


betweentwomidnights:
## Metal validation — the branch works, and the memory-only guess was too pessimistic

Ran on an M4 (32 GB, macOS 15) at `b58aded`. Adapter is the 5-step medium dora-rows rank/alpha-16
run from `train-runs/metal-medium-ratatat-512-tiled-5` — not a *good* adapter, but it shifts the
output hard (0.515 waveform cosine vs no-lora), so the path is genuinely exercised.

**`--lora` + `--encoding q4_k_m` works on Metal.** `lora: applied 1 adapter(s) [functional]`, clean
generation, base-norm sidecar written and reused. That ticks the Metal checkbox.

### Numbers

medium, 20 s, 8 steps, warm, base-norm cache primed. Peak RSS rather than VRAM — unified memory, so
RSS is the number that matters, and it covers the whole process (T5 + conditioner + SAME + DiT), not
just the DiT.

| backend | path | dit_compute | total | peak mem |
| --- | --- | --- | --- | --- |
| CUDA | f16 + merged | 699 ms | 5653 ms | 2994 MiB |
| CUDA | q4_k_m + functional | 728 ms | **3545 ms** | **1278 MiB** |
| Vulkan | f16 + merged | 523 ms | 6674 ms | 3049 MiB |
| Vulkan | q4_k_m + functional | 712 ms | **4310 ms** | **1248 MiB** |
| **Metal (M4)** | f16 + merged | 4925 ms | 13520 ms | 3071 MiB |
| **Metal (M4)** | q4_k_m + functional | 6340 ms | **11720 ms** | **1973 MiB** |

**13.3% faster end to end, 36% less memory.** Less speed than CUDA's 37% / Vulkan's 35% — expected,
since `docs/METAL.md` §9 already showed quantization is compute-flat on Metal, so there is no
quantization speedup underneath the load-time win. But it is a net gain, not the "memory-only or
small net loss" the checklist guessed.

Per-step cost of going functional, f16 base, like for like: **+21.8%** (4925 → 6000 ms). That sits
between CUDA's +4% and Vulkan's +36%.

### Functional vs merged agree almost exactly on Metal

Same f16 base, same adapter, `SA3_FUNCTIONAL_LORA=1` vs default merge:

| backend | functional vs merged, f16 |
| --- | --- |
| CUDA | 0.991400 |
| Vulkan | 0.950966 |
| **Metal** | **0.999711** (envelope 0.999990) |

Metal is by far the tightest of the three. Worth noting because it is decent evidence the residual
formulation itself is right — where the two paths diverge on Vulkan, that looks like Vulkan being
loose between any two paths rather than the math being off.

### The one that might change a default

**At 8 steps the functional path beats merging on an f16 base too** — 12.18 s vs 13.52 s. Merging
is not free on Metal: it costs ~2.4 s up front, and at 8 steps that outweighs the accumulated
per-step residual cost.

| steps | f16 merged | f16 functional |
| ---: | ---: | ---: |
| 8 | 13.52 s | **12.18 s** |
| 32 | 29.30 s | 29.97 s |

Crossover is somewhere around 20–25 steps. So on Metal, "f16 bases merge by default because the
merge costs nothing per step" is inverted for the common 8-step case. Not proposing a default change
from one adapter type on one machine — `dora-rows` merging also pays for the dora row work, and plain
`lora` may merge much more cheaply — but it is worth measuring on CUDA/Vulkan before the current
heuristic is treated as settled.

### Method note

First pass at these numbers was contaminated: running configs back-to-back let later ones inherit a
hot machine, which inflated 32-step functional to 36.29 s. Re-run interleaved with a 40 s cooldown
before every generation it is 29.97 s. Everything above is from the interleaved run, best of 3, with
run-to-run spread under 0.07 s. Flagging it because this M4 throttles enough over a long benchmark
sequence to invent a result that is not there.

### Small things

- `SA3_FUNCTIONAL_LORA` is a bare `getenv` presence test (`src/sa3_pipeline.h:998`), so
  `SA3_FUNCTIONAL_LORA=0` *enables* it. Fine as-is, just a sharp edge if it ever reaches a user-facing
  config.
- Untested here, still open from the checklist: multi-adapter composition, adapter families beyond
  `lora`/`dora-rows`.

Issue #22, opened 2026-07-20
web app interface update , Q4_KM compat layer and trainer web interface


pillopaus-project:
those are  based on your main  sa3.cpp  version.

new branch at : 
https://github.com/pillopaus-project/sa3.cpp/tree/q4km-webapp-fixes
for Q4_KM quantized models compatibilty ( incorporates web app ui).

## Q4_KM quantization (ported from sa3.cpp-reworked)
- New tools/sa3-quantize.cpp, sa3-quant-check.cpp, sa3-quant-eval.cpp; added CMake targets.
- ggml/src/ggml-cuda/getrows.cu: added k_get_rows_k kernel + Q4_K/Q5_K/Q6_K dispatch  (dequantize_q4_K_elem/q5_K_elem/q6_K_elem) so Q6_K embed.weight loads on CUDA via get_rows.
- ggml/src/ggml-cuda/dequantize.cuh: moved get_scale_min_k4 here (removed from convert.cu).
- ggml/src/gguf.cpp + gguf.h: added gguf_get_tensor_ndims / gguf_get_tensor_ne accessors.
- src/sa3_pipeline.h: ModelPaths::resolve now auto-detects encoding candidates  {F16,F32,Q4_KM,Q8_0} for DiT/SAME.
- tools/sa3-generate.cpp: --encoding help updated to f16|f32|q4_km|q8_0 (auto).
- Verified: quant-check all 85 tensors cos>=0.99; quant-eval T5 F32 vs Q4_KM CPU 0.909,  CUDA 0.911 (get_rows fix works).

## Webapp fixes
- Reload-after-generate bug: pushPastSong now uses savePastSongs() (try/catch + cap at 12,  trim-and-retry) and generate()/generateLoop() wrap the archive call so a localStorage  QuotaExceededError can no longer block generation or stick the button disabled.
- Loaded-models display: /health now returns actual_encoding (real encoding of resolved
  DiT/SAME files) via detect_encoding() + g_actual_encoding in ensure_loaded(); web header  shows model / actual_encoding.
- Keep Models Resident checkbox (was a no-op): server now frees g_pipe + g_loaded=false  after generation when keep_models==false (default frugal); keep_models==true keeps it  resident. Verified both directions on CPU.
- Regenerated src/embedded_web.h; web/app.ts + web/app.js kept in sync; npx tsc --noEmit 0  errors; sa3-server rebuilt (CUDA).

pillopaus-project:
[...]

PR #25, opened 2026-07-25
server: serve a front-end directory with --web-dir

betweentwomidnights:
First of the two changes from #22 aimed at getting @pillopaus-project's web interface to
zero changes in our C++. Deliberately small and standalone so it can be argued with before
it lands.

## The problem

Today a front-end has to be compiled into the binary (`embedded_web.h`, regenerated by a
script) and reach the network by patching `sa3-server.cpp` to add a route per asset. That
means every UI edit regenerates a header, and every one of our changes to that file is a
rebase for whoever maintains the UI.

## What this does

```sh
sa3-server --web-dir ./web          # or SA3_WEB_DIR=./web
```

Serves that directory at `/`. `GET /` resolves to `index.html`, `GET /app.js` to
`web/app.js`, and so on. Off by default — without the flag the server is API only and
`GET /` is a 404, exactly as before.

The serving is httplib's own file handler, not hand-rolled: it rejects `..`, backslashes
and nulls in the path, then canonicalizes the result and re-checks it against the base
directory so a symlink or junction cannot escape either. It also supplies `ETag` /
`Last-Modified` and directory redirects.

## Two rough edges, surfaced rather than hidden

- **httplib matches static files before route handlers, and only for GET/HEAD.** A file
  named `health`, `loras` or `prompts` in the web dir would shadow that GET endpoint. The
  server warns at startup if it finds one. POST routes (`/generate`, `/generate/loop`,
  `/unload`) can never be shadowed. I chose warn-not-refuse because it is the caller's
  directory and the collision is legal, just almost certainly a mistake — happy to make it
  fatal if that reads better.
- **A `--web-dir` that is not a directory exits non-zero** rather than starting up and
  quietly serving nothing. A missing `index.html` is only a note, so assets can be built
  into place after launch.

## Verified

| case | result |
| --- | --- |
| `GET /`, `/app.js`, `/sub/style.css` | 200, correct bodies |
| `GET /health` with web dir mounted | 200, still the API |
| `GET /nope.js` | 404 |
| `/../secret`, `/../../secret`, `/sub/../../secret` | 404, nothing leaked |
| `/%2e%2e/secret`, `/..%2fsecret`, `/....//secret` | 404, nothing leaked |
| `/..\secret`, `/sub\..\..\secret` | 404, nothing leaked |
| file named `health` in web dir | warned at startup, and does shadow |
| `--web-dir` at a non-directory | exits 1 with a clear message |
| no `--web-dir` | `web=(none)`, `GET /` 404, `GET /health` 200 |

## What this does not cover

@pillopaus-project — this replaces the `/` and `/app.js` routes in your `web-app` branch and
retires `embedded_web.h` plus `gen_embedded_web.py`, but it does **not** cover `/init-audio`
and `/init-audio/upload`. Those are a real API feature, not static files, so static serving
cannot absorb them; upstreaming them is a separate piece and I would rather take your version
than reinvent it.

Also worth flagging: your `web-app` branch rebases onto current `main` completely clean — I
trial-merged it, zero conflicts — so nothing here blocks you in the meantime.

`SA3_WEB_DIR` follows the existing env+flag convention the other directories use. The server
config work is the other half of #22 and will fold all of them into one place; this keeps the
surface consistent until then rather than inventing a second pattern.

### DIFF
--- a/docs/SERVER.md
+++ b/docs/SERVER.md
@@ -16,6 +16,7 @@ for progress and, on completion, the base64 audio. That makes it a drop-in for a
 #       --threads N (or SA3_THREADS, CPU backend only)
 #       --prompts-dir DIR (or SA3_PROMPTS_DIR)
 #       --source-loras-dir DIR (or SA3_SOURCE_LORAS_DIR)
+#       --web-dir DIR (or SA3_WEB_DIR) — serve a front-end at / ; omit for API only
 ```
 
 on Windows, after `.\build.cmd cuda`, `server.cmd` picks the built backend and keeps the server in the terminal (close it or Ctrl+C to stop). extra args pass through:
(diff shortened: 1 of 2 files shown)


betweentwomidnights:
Closing for now — want to validate against the ableton-extension/gary4juce (a downstream consumer of this server) before proposing it. Branch stays up; will reopen once that's checked.

update: nothing's broken downstream for me, so this feels ready.

betweentwomidnights:
@pillopaus-project when you get a chance, this is the first half of getting your web interface off patching sa3-server.cpp. `--web-dir` serves a directory at / so index.html and app.js are just files, which means embedded_web.h and gen_embedded_web.py go away entirely and you stop regenerating a header every time you touch the UI. Off by default, so nothing changes for anyone not using it.

Two things worth your opinion:

1. httplib matches static files before route handlers, so a file named health/loras/prompts, or a directory named poll_status, would shadow that GET endpoint. Right now the server just warns at startup. I could make it refuse to start instead. poll_status is the one that actually matters since clients poll it for progress.
2. I went with the existing env+flag convention (SA3_WEB_DIR) to stay consistent, knowing the config work will fold all of these into one place later.

/init-audio and /init-audio/upload are next, in their own PR. Static serving can't cover those since they're a real API, and I'd rather start from your version than reinvent it. One thing I want to talk through there: the upload handler takes the filename straight from the multipart header and joins it to the audio-in dir, and std::filesystem replaces the whole path if that filename is absolute, so a request can write outside the directory. Easy to fix, just wanted to raise it in its own thread rather than bury it here.

Also your web-app branch rebases onto current main completely clean, I checked, so none of this blocks you in the meantime.

pillopaus-project:
dont worry, its your app so its up to you how/when/if to implement feats.
Im still frankly surprised from the Q4 vs F16 speed gain... reduced memory footprint was the target but it was supposed to have some overhead from the dequant so i was expecting a small if any speed loss. 
keep up the nice work.
PP

betweentwomidnights:
yes I was surprised to find i may not even need to retrain LoRAs for use with q4 as well...my existing ones seem to apply nicely. although i'm told it will probably be more efficient if i just train some on a q4 base. the hardware implications have me excited

PR #26, opened 2026-07-26
server: init-audio pool for browser clients

betweentwomidnights:
Second half of #22. `/generate` takes `init_path`, a filesystem path, which is fine for a DAW
plug-in but impossible for a browser. These two are that bridge, ported from your `web-app`
branch:

```
GET  /init-audio         lists the pool
POST /init-audio/upload  puts a wav in it, returns the path to pass as init_path
```

The listing is the same directory scan `/loras` already does. I did rewrite the upload: it joined
the multipart filename straight onto the pool dir, and `fs::path(dir) / name` throws away `dir`
when `name` is absolute, so a filename of `C:/Windows/Temp/x.exe` wrote there. Now it takes bare
`.wav` names only (rejected rather than quietly renamed), caps at 256 MiB streamed to disk, and
warns at startup if the server is bound off loopback.

One open question in the comment below — it depends on how you serve the UI, so I didn't want to
guess.

### DIFF
--- a/docs/SERVER.md
+++ b/docs/SERVER.md
@@ -17,6 +17,7 @@ for progress and, on completion, the base64 audio. That makes it a drop-in for a
 #       --prompts-dir DIR (or SA3_PROMPTS_DIR)
 #       --source-loras-dir DIR (or SA3_SOURCE_LORAS_DIR)
 #       --web-dir DIR (or SA3_WEB_DIR) — serve a front-end at / ; omit for API only
+#       --audio-in-dir DIR (or SA3_AUDIO_IN_DIR) — init-audio pool for browser clients
 ```
 
 on Windows, after `.\build.cmd cuda`, `server.cmd` picks the built backend and keeps the server in the terminal (close it or Ctrl+C to stop). extra args pass through:
@@ -38,6 +39,8 @@ it binds to `127.0.0.1` by default (local only). The model loads lazily on the f
 | `POST` | `/generate/loop` | same request plus `bpm`/prompt BPM and `bars` for exact-length loop generation |
 | `GET`  | `/poll_status/<session_id>` | `{success, generation_in_progress, progress, step, total_steps, status, queue_status, ...}`; on `status:"completed"` also `audio_data` (base64 wav) + `meta:{seed}` |
 | `POST` | `/unload`   | frees the model (full VRAM release) → `{status:"unloaded"}` |
+| `GET`  | `/init-audio` | `{success, files:[{name,path,bytes}], audio_in_dir, max_upload_bytes}` — the init-audio pool |
+| `POST` | `/init-audio/upload` | multipart `file=` → `{success, name, path, bytes}`; the returned `path` is what you pass as `init_path` |
 
 `status` runs `queued → generating → encoding → completed` (or `failed`); `progress` is `0..100`
 (sampling `0→90`, decode `→100`). poll until `status == "completed"`, then base64-decode `audio_data`.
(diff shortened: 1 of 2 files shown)


betweentwomidnights:
Added a startup warning when the server binds off loopback, since the upload writes to disk and the 127.0.0.1 default is the only thing keeping it off the network.

One question I did not want to answer for you: a loopback bind does not stop a browser, since any page you have open can POST multipart to 127.0.0.1 cross-origin without a preflight. Checking the Origin header would block that, and native clients like the ableton extension send no Origin so they would be unaffected. But if you ever serve the UI from a dev server or from file://, a strict check would reject you and look like our bug. How do you run it while developing? If it is always served from sa3-server I will add it.

pillopaus-project:
i was busy working on a hw port of  a digital oscillator , sorry for the late reply. go on as you wish, dont worry, im anyway not using sa3 , as his only point of interest for me was the loop creation function , but  looks like it doesnt really follows bpms and to generate complete songs, acestep-xl simply runs circles around sa3-medium..
have a nice week.

betweentwomidnights:
interesting i get fantastic results with the loop generation using pytorch. can even do odd bpms like 103. admittedly haven't messed much with loops yet in the c++ but im not sure why that'd change. i use both models for various tasks but i also use the DAW so alot of pain points with both i just dont experience in the same way as others

pillopaus-project:
I ve tried more extreme bpm differences (60/160/35/190) as its more revealing and i never found something slower than 109 or faster than 136/138( dont remember the exact one).
interesting the pythorch code works... i should have a look at that one of these days...
PP

betweentwomidnights:
yeah i wouldn't expect many models to be good at anything below 60 or much higher than 172 lol that's just a dataset problem. in reality, a loop at 70bpm is going to line up with whatever you're making at 35 no? I mostly use loras, but they do just fine producing loops at 84/93/72 etc bpm for me. Using the ableton extension with the c++ i see them lining up on the grid just fine so I suppose I have tested it. 

My bpm labelling isn't anything crazy so the base model is definitely capable of it on its own. I've been under the impression the LoRAs don't really affect its bpm/key awareness (it's quite good at key/scale)

pillopaus-project:
the 35/180 are just extreme examples to check the model knowledge of the bpm .
Acestep understands 70 or 150 quite well... sa3.cpp for me never moves from +-110 to +- 135. i ve never tried the original python engine

betweentwomidnights:
odd. yeah i find both models great at bpm i am not sure what you're doing with your prompts. Just produced 70 bpm fine with base model in sa3.cpp. it's such that i can abstract away the bpm entirely and attach it to the DAW's global bpm in my applications. with odd bpm's there will be slight drift with both ace step and sa3 but neither is drifting more than the other.

the key difference in the two is prompt length. if you are using the length of prompts ace step expects they are probably getting truncated. sa3 has a 256 token cap (99% sure but i'll have to check how we're handling it here). just a thought that if you're putting bpm at the end of a very long prompt that could be why.  sa3 likes them very short and sweet, and when using a blend of multiple DoRA adapters, you actually don't even need a prompt at that point. just bpm and key.

pillopaus-project:
i dont put bpm in the prompt, but in the bpm part of the loop interface... im used to acestep NOT  liking bpm in the captions so ive been using the same kind of prompt... the examples were made with a damn easy captioning : " deep house groove loop with upbeat bassline and analog synth pad chords"  or similar style prompts, varying the musical style so i dont really think the problem's the captioning too long or too difficult.
PP

betweentwomidnights:
interesting. it could be i need to take a look at the loop api for sa3.cpp then. yes, in ace-step bpm should be used as part of the metas and not part of the caption. in sa3 it should be part of the prompt. So in interfaces all we do is just append it with a comma. Same for key. btw I find I don't need to tell it 'loop'. the trimming is going to make it a loop regardless, and the padding tail seconds should be 6 so that it generates longer than the desired duration for the intended effect. busy with some other stuff right now but I'll take a look and try to figure out why you're struggling and I have such great results in my interfaces.

pillopaus-project:
i didnt try in the last 10/15 days , honestly; i marked it as "not good" back then and never tried again. will try in the next days if something changed lately.
PP

betweentwomidnights:
it could also be the fact that putting 'house' in the prompt is conflicting with the bpm you want since house is a genre that has a specific bpm range. That would be something it learned from the data. 

i did just find the outputs i got asking it for house groove at 70 bpm lined up but idk if i would call it house at all lol, which makes sense from a dataset perspective...idk i can't stand base model outputs or prompting in general, so to me this model really is 1000% less enjoyable without using multi-lora blends. Then all you're doing is locking the seed, appending bpm/key and adjusting sliders. I've even trained an adapter on a single 12 second arpeggio synth clip so that it works as an arpeggiator effect to apply to my other adapters. It's a very different experience from the usual ai music model.

pillopaus-project:
EDIT : 
I just tried with the bpm in the caption " acid jazz loop at 60 bpm" or "hardcore techno loop at 160 bpm" without touching the "120" bpm default and indeed it follows bpm asked quite good.. 
Its then something related to the bpm injection on the /loop endpoint.. 
wil have a look at the web interface before the server part.
thanks for pointing out it worked for you with captioning. 

with acestep ive a library of timbre reference VAE that i created that   (when the the lyrics structure defining part are right and im happy with the end track) I apply to the same seed  so i get  different versions  of the same track and whats more :"consistent between songs"  instrumentations,like a band  or like when i compose music using my hw synths.. the sonic imprint is coherent and gives a distinct "artist" style.
It's like swapping the instruments in a band session or the synthracks in an electronic setup  , minus the hassle of recabling and  setting up the patches  ;)

PP

PR #28, opened 2026-08-03
train: all eight adapter families train, and 25x faster

betweentwomidnights:
Two of the eight adapter families could not be trained at all, and six of the eight trained
tens of times slower than they should have. Both are fixed here, and neither cause was what it
looked like.

## dora-cols could not build a backward graph

`dora-cols` and `dora-cols-xs` aborted in `ggml_build_backward_expand` on
`GGML_ASSERT(ggml_is_padded_1d(a))`.

The obvious reading is that column normalization is at fault — both failing families normalize
over columns, and no other family does. That reading is wrong, and `bora` proves it: `bora` calls
the very same `train_lora_apply_col_norm` and never crashed.

The assert lives in `ggml_scale`, which requires a padded-1d input. ggml differentiates transpose
as a bare `ggml_transpose(grad)` — a non-contiguous view, with no `cont` (`ggml.c:6756`), unlike
`GGML_OP_CONT` which explicitly re-materializes (`ggml.c:6701`). The column-norm helper transposes
`v`, so the gradient arriving back at `v` is that view. For `dora-cols`, `v = w + scale(delta)`,
and add's backward forwards its gradient unchanged straight into the `ggml_scale` node, whose own
backward calls `ggml_scale` on the view.

So the real trigger is a transposed gradient reaching a scale node. `bora` survives only because
its `rms_norm`/`mul` layer sits between the transpose and the scale and allocates a fresh
contiguous gradient on the way past; `dora-rows` never transposes at all.

The fix interposes a `cont` on the helper's input. In the forward pass it is a copy; the point is
the backward, where CONT's backward makes the gradient contiguous before it can reach the scale.
`bora` is bit-identical before and after.

## The other six were paying for 24.6 GiB of VRAM they did not have

At the reference 512 frames, `dora-cols` measured **92.76 s/step** against `dora-rows`' steady
1.25 s. That looks like an arithmetic problem. It is not:

| | graph | allocator |
|---|---|---|
| `dora-rows` | checkpointed, 24 block graphs | **407 MiB** |
| `dora-cols` | one monolithic graph | **24,629 MiB** |

24.6 GiB requested on an 8,150 MiB card, so the driver pages it to system RAM and every kernel
crawls. These families were never slow — they were never checkpointed.

Checkpointing was gated to `lora`/`dora-rows` for one reason: it needs the adapter to reach
`dit_lin` inside each per-block graph, and the other families install a materialized `W_eff` on
`dit.overrides` built once for the whole step. That is a plumbing coupling, not a mathematical
one.

Now each segment builds the `W_eff` it consumes. `F` is forward-only, so gallocr reuses each
one's buffer as soon as its block has run; `T`, `B_l` and `H` only ever build the targets their
own segment owns. Peak stays one segment's working set, which is the entire point of the
checkpointed decomposition.

## Parity

The arithmetic is untouched and the numbers confirm it. Every family reproduces its monolithic
loss and `grad_norm` **to the last digit**, at 128 and 512 frames. `dora-cols` at 512 frames:
`0.653433 / 0.855772` and gnorm `0.1077 / 0.1237` on both paths, with the allocator down from
24,629 to 1,126 MiB and the step from 92.76 s to 1.95 s.

## Results

Steady-state s/step, medium, rank 16, 512 frames, CUDA / RTX 5070 Laptop, averaged over steps
10–40:

| family | s/step | first step |
|---|---:|---:|
| `lora` | 0.80 | 1.02 |
| `lora-xs` | 1.05 | 1.33 |
| `dora-rows` | 1.25 | 6.17 |
| `dora-rows-xs` | 1.57 | 2.00 |
| `dora-cols-xs` | 1.84 | 2.31 |
| `dora-cols` | 1.95 | 2.35 |
| `bora-xs` | 2.33 | 2.67 |
| `bora` | 2.47 | 2.88 |

The ordering matches the arithmetic: plain, then the `-xs` variants, then row norm, then column
norm, then `bora` doing both. End to end, matched 2000-step runs on the same dataset came in at
34 min (`dora-rows`) and 64 min (`dora-cols`) — `dora-cols` was previously a ~51 hour proposition.

Measure over enough steps to reach steady state: `dora-rows` pays a one-off 6.17 s `base_norm_sq`
reduction on step 1 against a 1.25 s steady step, so short averages over-report it ~3x and rank
the families wrongly. An earlier revision of this PR did exactly that and claimed the functional
path was the slowest of the eight; it is third fastest, and that claim is withdrawn.

## Verification

- 17/17 ctest pass.
- All 8 families train at 128 and 512 frames, exact parity against the monolithic reference.
- Matched 2000-step `dora-rows` vs `dora-cols` runs on a real dataset both converge
  (loss 0.6727 → 0.6098 and 0.6742 → 0.6259 over first/last 100 steps) and produce audio.
- Not yet run on Metal — the checkpointed path is the one that changed, so it is worth
  confirming there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,9 +134,9 @@ what's next:
 > dora-cols and the -xs variants are formula-validated only (no trained checkpoint to a/b yet) but
 > share the same apply path.
 >
-> **training an adapter here** covers all eight, at 0.5–3.3 s/step on a laptop 5070 (medium, rank
-> 16, 128 frames) — roughly one to two hours for a 2000-step run at the reference 512 frames. see
-> [docs/TRAINING.md](docs/TRAINING.md).
+> **training an adapter here** covers all eight, at 0.8–2.5 s/step on a laptop 5070 (medium, rank
+> 16, the reference 512 frames) — a 2000-step run measured 34 min for dora-rows and 64 min for
+> dora-cols. see [docs/TRAINING.md](docs/TRAINING.md).
 
 credits:
 
--- a/README.md
+++ b/README.md
@@ -134,9 +134,9 @@ what's next:
 > dora-cols and the -xs variants are formula-validated only (no trained checkpoint to a/b yet) but
 > share the same apply path.
 >
-> **training an adapter here** covers six of the eight. dora-cols and dora-cols-xs currently abort in
-> the backward pass, and the bora families train correctly but ~110x slower per step than lora, so a
-> full run takes over a day. see [docs/TRAINING.md](docs/TRAINING.md).
+> **training an adapter here** covers all eight, at 0.5–3.3 s/step on a laptop 5070 (medium, rank
+> 16, 128 frames) — roughly one to two hours for a 2000-step run at the reference 512 frames. see
+> [docs/TRAINING.md](docs/TRAINING.md).
 
 credits:
 
--- a/docs/TRAINING.md
+++ b/docs/TRAINING.md
@@ -169,28 +169,37 @@ pass `--svd-bases bases.gguf` to load precomputed bases instead — generate the
 `python tools/compute_svd_bases.py --dit models/... --rank 8 --out bases.gguf` for exact
 `torch.linalg.svd` parity with the reference implementation.
 
-Accepted is not the same as working. Six of the eight train today; the two column-normalizing
-families do not:
+All eight train, and all eight are within a small factor of each other per step:
 
-| adapter type | trains | s/step |
-|---|---|---|
-| `lora` | yes | 0.5 |
-| `dora-rows` (default) | yes | 3.2 |
-| `lora-xs` | yes | 16.3 |
-| `dora-rows-xs` | yes | 30.9 |
-| `bora` | yes | 59.4 |
-| `bora-xs` | yes | 69.6 |
-| `dora-cols` | **no** | — |
-| `dora-cols-xs` | **no** | — |
+| adapter type | s/step (128 frames) |
+|---|---|
+| `lora` | 0.48 |
+| `lora-xs` | 0.71 |
+| `dora-rows-xs` | 1.24 |
+| `dora-cols-xs` | 1.42 |
+| `dora-cols` | 1.61 |
+| `bora-xs` | 1.97 |
+| `bora` | 2.17 |
+| `dora-rows` (default) | 3.30 |
 
-`dora-cols` and `dora-cols-xs` abort during the backward pass with
-`GGML_ASSERT(ggml_is_padded_1d(a))` in `ggml_build_backward_expand`, which correlates with the
-column normalization those two share.
+CUDA / RTX 5070 Laptop, medium, rank 16 — ratios rather than absolutes. At the reference 512
+frames a 2000-step run is roughly one to two hours for any family.
 
-Timings are CUDA / RTX 5070 Laptop, medium, 128 frames, and are meant as ratios rather than
-absolutes. The spread matters when planning a run: `dora-rows` at 3.2 s/step is a couple of hours
-for 2000 steps, while `bora-xs` at 69.6 s/step is closer to a day and a half for the same
-schedule. `dora-rows` is the default and by far the best-trodden path.
+Two things used to make this table much worse, both fixed:
+
+- `dora-cols` and `dora-cols-xs` aborted during the backward pass on
+  `GGML_ASSERT(ggml_is_padded_1d(a))`. ggml differentiates transpose into a non-contiguous view
+  and `ggml_scale` rejects one, so the column-norm helper's transposed gradient reached the
+  scale on the low-rank delta. It now interposes a `cont`, whose backward re-materializes the
+  gradient contiguously.
+- Six of the eight ran an unbatched monolithic backward asking for 24.6 GiB of allocator on an
+  8 GiB card, so the driver paged it to system RAM: `dora-cols` measured 92.76 s/step at 512
+  frames against 2.05 now. Gradient checkpointing had been gated to the two families
+  `dit_lin` can apply functionally; the rest now build their effective weight inside whichever
+  segment graph consumes it, so they are checkpointed too. See `--checkpoint-backward`.
+
+`dora-rows` remains the default and the best-trodden path, and is now the slowest of the eight —
+the functional application costs more per step than materialized-plus-checkpointed at this size.
 
 Each sample is conditioned by its caption text. A dataset-level `prompt_config.json` can optionally
 compose that caption with general metadata tags or path-derived text.
--- a/docs/TRAINING.md
+++ b/docs/TRAINING.md
@@ -208,6 +208,35 @@ Two things used to make this table much worse, both fixed:
 Each sample is conditioned by its caption text. A dataset-level `prompt_config.json` can optionally
 compose that caption with general metadata tags or path-derived text.
 
+### Path-derived captions are off by default
+
+`use_paths` defaults to **false**, which is a deliberate departure from underfit, where filenames
+are part of the dice pool. Turn it on only if your filenames actually read as descriptions of the
+music:
+
+```json
+{ "prompt_config": { "use_tags": true, "use_paths": true, "balance": { "tags": 40, "paths": 30 } } }
+```
+
+With that on, roughly a third of steps train on the file path verbatim — and `path_hide_ext`,
+`path_hide_dirs` and `path_hide_topmost` all default to false, so a track lands in the model as
+
+```
+03 - Gettysburg [Q-gtqz_il-w].wav
+```
+
+A track number, a YouTube ID and a file extension are not a description of a piece of music, and
+training on them teaches the adapter to emit your material in response to noise. That dilutes the
+conditioning the rest of the run is trying to build, and it shows up at inference as an adapter
+that responds best to no prompt at all rather than to a prompt from its training distribution.
+
+If your filenames are genuinely descriptive, `use_paths` is worth having, and `hide_ext` /
+`hide_dirs` / `hide_topmost_dir` trim the parts that are not. Otherwise leave it off and let the
+tags plus the CFG dropout (`--cfg-dropout-prob`, default 0.1) do the generalizing.
+
+Check what a run is actually training on before committing hours to it — the per-step log prints
+the composed caption, and a quick tally of `prompt="..."` across the log will show the mix.
+
 ## Target Scopes
 
 Which DiT weights get an adapter. Every 2D `dit.*.weight` is eligible; a scope narrows that.
--- a/docs/TRAINING.md
+++ b/docs/TRAINING.md
@@ -169,21 +169,26 @@ pass `--svd-bases bases.gguf` to load precomputed bases instead — generate the
 `python tools/compute_svd_bases.py --dit models/... --rank 8 --out bases.gguf` for exact
 `torch.linalg.svd` parity with the reference implementation.
 
-All eight train, and all eight are within a small factor of each other per step:
+All eight train, and all eight are within about 3x of each other per step:
 
-| adapter type | s/step (128 frames) |
-|---|---|
-| `lora` | 0.48 |
-| `lora-xs` | 0.71 |
-| `dora-rows-xs` | 1.24 |
-| `dora-cols-xs` | 1.42 |
-| `dora-cols` | 1.61 |
-| `bora-xs` | 1.97 |
-| `bora` | 2.17 |
-| `dora-rows` (default) | 3.30 |
+| adapter type | s/step | first step |
+|---|---|---|
+| `lora` | 0.80 | 1.02 |
+| `lora-xs` | 1.05 | 1.33 |
+| `dora-rows` (default) | 1.25 | 6.17 |
+| `dora-rows-xs` | 1.57 | 2.00 |
+| `dora-cols-xs` | 1.84 | 2.31 |
+| `dora-cols` | 1.95 | 2.35 |
+| `bora-xs` | 2.33 | 2.67 |
+| `bora` | 2.47 | 2.88 |
 
-CUDA / RTX 5070 Laptop, medium, rank 16 — ratios rather than absolutes. At the reference 512
-frames a 2000-step run is roughly one to two hours for any family.
+CUDA / RTX 5070 Laptop, medium, rank 16, the reference 512 frames, averaged over steps 10-40. A
+2000-step run is 30-90 minutes depending on family; the two measured end to end came in at 34 min
+(`dora-rows`) and 64 min (`dora-cols`).
+
+Measure over enough steps to reach steady state. The first step carries graph build and, for
+`dora-rows`, the one-off `base_norm_sq` reduction over every targeted weight — 6.17 s against a
+1.25 s steady step, so a 2- or 3-step average is mostly setup and ranks the families wrongly.
 
 Two things used to make this table much worse, both fixed:
 
@@ -192,14 +197,13 @@ Two things used to make this table much worse, both fixed:
   and `ggml_scale` rejects one, so the column-norm helper's transposed gradient reached the
   scale on the low-rank delta. It now interposes a `cont`, whose backward re-materializes the
   gradient contiguously.
-- Six of the eight ran an unbatched monolithic backward asking for 24.6 GiB of allocator on an
-  8 GiB card, so the driver paged it to system RAM: `dora-cols` measured 92.76 s/step at 512
-  frames against 2.05 now. Gradient checkpointing had been gated to the two families
-  `dit_lin` can apply functionally; the rest now build their effective weight inside whichever
-  segment graph consumes it, so they are checkpointed too. See `--checkpoint-backward`.
+- Six of the eight ran a monolithic backward asking for 24.6 GiB of allocator on an 8 GiB card,
+  so the driver paged it to system RAM: `dora-cols` measured 92.76 s/step at 512 frames against
+  1.95 now, with the allocator down to 1126 MiB. Gradient checkpointing had been gated to the two
+  families `dit_lin` can apply functionally; the rest now build their effective weight inside
+  whichever segment graph consumes it, so they are checkpointed too. See `--checkpoint-backward`.
 
-`dora-rows` remains the default and the best-trodden path, and is now the slowest of the eight —
-the functional application costs more per step than materialized-plus-checkpointed at this size.
+`dora-rows` remains the default and the best-trodden path.
 
 Each sample is conditioned by its caption text. A dataset-level `prompt_config.json` can optionally
 compose that caption with general metadata tags or path-derived text.
--- a/src/train_ckpt.h
+++ b/src/train_ckpt.h
@@ -37,6 +37,7 @@
 #include "ggml-backend.h"
 
 #include <cstdio>
+#include <map>
 #include <string>
 #include <vector>
 
@@ -66,6 +67,7 @@ struct TrainCkptBlock {
     ggml_tensor* grad_ctx = nullptr;    // this block's dL/dcontext contribution [dim, Ctx]
     ggml_tensor* grad_gcond = nullptr;  // this block's dL/dgcond contribution [6*dim]
     std::vector<size_t> param_idx;      // indices into lora.params / TrainDitCkpt::params
+    ggml_gallocr_t alloc = nullptr;     // borrowed from TrainDitCkpt::ballocs; see the note there
 };
 
 struct TrainDitCkpt {
@@ -102,7 +104,15 @@ struct TrainDitCkpt {
     ggml_tensor* velocity = nullptr;
     ggml_tensor* loss = nullptr;
     std::vector<size_t> tail_param_idx;
-    ggml_gallocr_t balloc = nullptr;    // shared by all block graphs (re-alloc'd per use)
+    // One allocator per DISTINCT block-graph shape, not one for all 24 blocks. With a uniform
+    // scope every block graph is identical, so a single allocator plans once and this is exactly
+    // the old behaviour. A scope that adapts only some blocks (--lora-include dit.0.) makes the
+    // graphs structurally different, and a shared gallocr then re-plans on every block boundary:
+    // ggml_gallocr_alloc_graph -> needs_realloc -> ggml_gallocr_reserve, which FREES and
+    // reallocates the buffer 24 times a step. That corrupted the step nondeterministically
+    // (~30% of runs died in ggml_cuda_set_device with "invalid device ordinal", or took an
+    // access violation). Grouping by shape means no allocator ever re-plans.
+    std::vector<ggml_gallocr_t> ballocs;
     std::vector<TrainCkptBlock> blocks;
     ggml_context* hctx = nullptr;       // H: head VJP, fwd+bwd
     ggml_gallocr_t halloc = nullptr;
@@ -119,7 +129,7 @@ struct TrainDitCkpt {
 inline void free_train_dit_ckpt(TrainDitCkpt& ck) {
     if (ck.falloc) ggml_gallocr_free(ck.falloc);
     if (ck.talloc) ggml_gallocr_free(ck.talloc);
-    if (ck.balloc) ggml_gallocr_free(ck.balloc);
+    for (ggml_gallocr_t a : ck.ballocs) if (a) ggml_gallocr_free(a);
     if (ck.halloc) ggml_gallocr_free(ck.halloc);
     for (TrainCkptBlock& b : ck.blocks) if (b.ctx) ggml_free(b.ctx);
     if (ck.fctx) ggml_free(ck.fctx);
@@ -372,10 +382,8 @@ inline bool build_train_dit_ckpt(GgufModel& dit, const DitConfig& dc, const Trai
             return fail("failed to allocate tail training graph");
     }
 
-    // --- B_l: per-block VJP fwd+bwd, shared allocator (re-planned per use in the runner) ---
+    // --- B_l: per-block VJP fwd+bwd, one allocator per distinct graph shape ---
     {
-        out.balloc = ggml_gallocr_new(buft);
-        if (!out.balloc) return fail("failed to create block graph allocator");
         out.blocks.resize((size_t)dc.depth);
         for (int l = 0; l < dc.depth; ++l) {
             TrainCkptBlock& B = out.blocks[(size_t)l];
@@ -400,6 +408,25 @@ inline bool build_train_dit_ckpt(GgufModel& dit, const DitConfig& dc, const Trai
             ggml_set_output(B.grad_ctx);
             ggml_set_output(B.grad_gcond);
             keep_param_grads(B.graph, B.param_idx);
+        }
+
+        // Group blocks by graph shape and give each group its own allocator, reserved once here.
+        // Two blocks share an allocator only when their graphs are structurally identical, so
+        // ggml_gallocr_alloc_graph in the runner never takes the realloc path. A uniform scope
+        // yields exactly one group, i.e. the previous single-allocator behaviour and memory use.
+        std::map<int, size_t> shape_to_alloc;   // graph node count -> index into ballocs
+        for (int l = 0; l < dc.depth; ++l) {
+            TrainCkptBlock& B = out.blocks[(size_t)l];
+            const int shape = ggml_graph_n_nodes(B.graph);
+            auto it = shape_to_alloc.find(shape);
+            if (it == shape_to_alloc.end()) {
+                ggml_gallocr_t a = ggml_gallocr_new(buft);
+                if (!a) return fail("failed to create block graph allocator");
+                out.ballocs.push_back(a);
+                if (!ggml_gallocr_reserve(a, B.graph)) return fail("failed to reserve block graph allocator");
+                it = shape_to_alloc.emplace(shape, out.ballocs.size() - 1).first;
+            }
+            B.alloc = out.ballocs[it->second];
         }
     }
 
--- a/src/train_ckpt.h
+++ b/src/train_ckpt.h
@@ -23,8 +23,9 @@
 // VRAM-resident; the price is one extra forward pass. All step inputs, adapter tensors and
 // boundary/grad carriers live in ONE persistent backend buffer that no gallocr ever touches
 // (this also retires the "re-upload base_norm_sq every step" workaround — constants uploaded at
-// build survive). Functional adapter families only (lora / dora-rows); others use the
-// monolithic path in train_dit.h.
+// build survive). All adapter families: lora / dora-rows reach dit_lin functionally, and the rest
+// materialize a W_eff inside whichever segment graph consumes it (see install_overrides), which is
+// the whole reason they fit — built once for the step they were ~24.6 GB of live f32 weights.
 #pragma once
 
 #include "dit.h"
@@ -88,7 +89,8 @@ struct TrainDitCkpt {
     ggml_tensor* Gctx_in = nullptr;     // [dim, ctx_len] summed dL/dcontext (uploaded before H)
     ggml_tensor* Ggcond_in = nullptr;   // [6*dim] summed dL/dgcond
     std::vector<TrainDitParamTensors> params;  // adapter tensors, indexed like lora.params
-    DitLora dl;
+    std::vector<TrainLoraGraphParam> gp;       // same order; feeds train_lora_effective_weight
+    DitLora dl;                                // functional families only; empty otherwise
 
     // --- graphs ---
     ggml_context* fctx = nullptr;       // F: forward + boundary copies
@@ -131,10 +133,6 @@ inline void free_train_dit_ckpt(TrainDitCkpt& ck) {
 inline bool build_train_dit_ckpt(GgufModel& dit, const DitConfig& dc, const TrainLoraState& lora,
                                  int frames, int cond_dim, int ctx_len,
                                  TrainDitCkpt& out, std::string& err, bool inpaint = false) {
-    if (lora.adapter_type != "lora" && lora.adapter_type != "dora-rows") {
-        err = "checkpointed backward supports functional families only (lora, dora-rows)";
-        return false;
-    }
     if (frames <= 0 || cond_dim <= 0 || ctx_len <= 0) {
         err = "invalid DiT training graph dimensions";
         return false;
@@ -150,11 +148,19 @@ inline bool build_train_dit_ckpt(GgufModel& dit, const DitConfig& dc, const Trai
     out.ctx_len = ctx_len;
     const int64_t S = (int64_t)dc.mem_tokens + frames;
     const bool dora = lora.adapter_type == "dora-rows";
+    // Families dit_lin can apply functionally; the rest materialize a W_eff per segment graph
+    // (see install_overrides). Both are checkpointed — the split is only about how the adapter
+    // reaches dit_lin, not about whether the step fits in VRAM.
+    const bool functional = lora.adapter_type == "lora" || lora.adapter_type == "dora-rows";
+    const bool xs = lora.adapter_type.size() >= 3 &&
+                    lora.adapter_type.compare(lora.adapter_type.size() - 3, 3, "-xs") == 0;
     auto fail = [&](const std::string& msg) { err = msg; free_train_dit_ckpt(out); return false; };
 
     // --- persistent tensors ---
     {
-        const size_t n_tensors = 24 + (size_t)(dc.depth + 1) + lora.params.size() * 4;
+        // Per target, the widest family is bora-xs: U, V, M_xs, magnitude_r, magnitude_c. Standard
+        // families use at most lora_A, lora_B, magnitude and base_norm_sq.
+        const size_t n_tensors = 24 + (size_t)(dc.depth + 1) + lora.params.size() * 6;
         ggml_init_params ip = { ggml_tensor_overhead() * (n_tensors + 64), nullptr, true };
         out.pctx = ggml_init(ip);
         if (!out.pctx) return fail("ggml_init failed for persistent training tensors");
@@ -186,26 +192,56 @@ inline bool build_train_dit_ckpt(GgufModel& dit, const DitConfig& dc, const Trai
         for (const TrainLoraParam& hp : lora.params) {
             TrainDitParamTensors tp;
             tp.stem = hp.target.stem;
-            tp.lora_A = ggml_new_tensor_2d(out.pctx, GGML_TYPE_F32, hp.target.in, lora.rank);
-            tp.lora_B = ggml_new_tensor_2d(out.pctx, GGML_TYPE_F32, lora.rank, hp.target.out);
-            ggml_set_param(tp.lora_A);
-            ggml_set_param(tp.lora_B);
-            DitLoraParam dp;
-            dp.A = tp.lora_A;
-            dp.B = tp.lora_B;
-            dp.scale = lora.alpha / (float)lora.rank;
-            dp.in = hp.target.in;
-            if (dora) {
+            TrainLoraGraphParam gp;
+            if (xs) {
+                // Frozen SVD bases are inputs (uploaded, never trained); only M_xs is a parameter.
+                tp.U = ggml_new_tensor_2d(out.pctx, GGML_TYPE_F32, lora.rank, hp.target.out);
+                tp.V = ggml_new_tensor_2d(out.pctx, GGML_TYPE_F32, lora.rank, hp.target.in);
+                tp.M_xs = ggml_new_tensor_2d(out.pctx, GGML_TYPE_F32, lora.rank, lora.rank);
+                ggml_set_param(tp.M_xs);
+                gp.U = tp.U;
+                gp.V = tp.V;
+                gp.M_xs = tp.M_xs;
+            } else {
+                tp.lora_A = ggml_new_tensor_2d(out.pctx, GGML_TYPE_F32, hp.target.in, lora.rank);
+                tp.lora_B = ggml_new_tensor_2d(out.pctx, GGML_TYPE_F32, lora.rank, hp.target.out);
+                ggml_set_param(tp.lora_A);
+                ggml_set_param(tp.lora_B);
+                gp.lora_A = tp.lora_A;
+                gp.lora_B = tp.lora_B;
+            }
+            if (!hp.magnitude.empty()) {
                 tp.magnitude = ggml_new_tensor_1d(out.pctx, GGML_TYPE_F32, (int64_t)hp.magnitude.size());
                 ggml_set_param(tp.magnitude);
-                tp.base_norm_sq = ggml_new_tensor_1d(out.pctx, GGML_TYPE_F32, hp.target.out);
-                tp.base_norm_sq_host = train_row_norm_sq(dit.get(hp.target.weight_name), hp.target.in,
-                                                         hp.target.out, 1e-12f);
-                dp.dora = true;
-                dp.magnitude = tp.magnitude;
-                dp.base_norm_sq = tp.base_norm_sq;
+                gp.magnitude = tp.magnitude;
             }
-            out.dl[hp.target.weight_name] = dp;
+            if (!hp.magnitude_r.empty()) {
+                tp.magnitude_r = ggml_new_tensor_1d(out.pctx, GGML_TYPE_F32, (int64_t)hp.magnitude_r.size());
+                ggml_set_param(tp.magnitude_r);
+                gp.magnitude_r = tp.magnitude_r;
+            }
+            if (!hp.magnitude_c.empty()) {
+                tp.magnitude_c = ggml_new_tensor_1d(out.pctx, GGML_TYPE_F32, (int64_t)hp.magnitude_c.size());
+                ggml_set_param(tp.magnitude_c);
+                gp.magnitude_c = tp.magnitude_c;
+            }
+            if (functional) {
+                DitLoraParam dp;
+                dp.A = tp.lora_A;
+                dp.B = tp.lora_B;
+                dp.scale = lora.alpha / (float)lora.rank;
+                dp.in = hp.target.in;
+                if (dora) {
+                    tp.base_norm_sq = ggml_new_tensor_1d(out.pctx, GGML_TYPE_F32, hp.target.out);
+                    tp.base_norm_sq_host = train_row_norm_sq(dit.tensors.at(hp.target.weight_name),
+                                                             hp.target.in, hp.target.out, 1e-12f);
+                    dp.dora = true;
+                    dp.magnitude = tp.magnitude;
+                    dp.base_norm_sq = tp.base_norm_sq;
+                }
+                out.dl[hp.target.weight_name] = dp;
+            }
+            out.gp.push_back(gp);
             out.params.push_back(tp);
         }
 
@@ -235,6 +271,10 @@ inline bool build_train_dit_ckpt(GgufModel& dit, const DitConfig& dc, const Trai
     }
 
     const ggml_backend_buffer_type_t buft = ggml_backend_get_default_buffer_type(dit.backend);
+    // A materialized W_eff adds ~10 nodes per target on top of what the segment already needs, and
+    // bora-xs (xs delta + row norm + col norm) is the worst case, so the non-functional families
+    // get a larger node budget in every graph.
+    const size_t gsz = functional ? 1 : 2;
     auto graph_ctx = [&](size_t graph_size, bool grads) -> ggml_context* {
         ggml_init_params ip = { ggml_tensor_overhead() * graph_size * 3 +
                                 ggml_graph_overhead_custom(graph_size, grads), nullptr, true };
@@ -245,7 +285,8 @@ inline bool build_train_dit_ckpt(GgufModel& dit, const DitConfig& dc, const Trai
     auto keep_param_grads = [&](ggml_cgraph* g, const std::vector<size_t>& idx) {
         for (size_t i : idx) {
             const TrainDitParamTensors& tp = out.params[i];
-            for (ggml_tensor* p : { tp.lora_A, tp.lora_B, tp.magnitude }) {
+            for (ggml_tensor* p : { tp.lora_A, tp.lora_B, tp.M_xs, tp.magnitude,
+                                    tp.magnitude_r, tp.magnitude_c }) {
                 if (!p) continue;
                 ggml_tensor* gr = ggml_graph_get_grad(g, p);
                 if (gr) ggml_set_output(gr);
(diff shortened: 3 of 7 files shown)


betweentwomidnights:
## Metal validation — all eight train, and the M4 explains the silent half of the bug

Ran on an M4 (32 GB, macOS 15) at `35a48e3`. Medium, rank 16, 128 frames, ratatat latents.
All eight families train with real losses and gradients; `dora-cols` and `dora-cols-xs` are fixed here.

### The M4 saw a different symptom of the same bug

Before this branch, `bora`, `bora-xs` and `dora-rows-xs` on Metal did not merely run slowly — they
reported `loss: 0, grad_norm: 0` on every step, and did so *fast*. The logs give the reason:

```
[train] DiT fwd+bwd graph: 17726 nodes, gallocr buffer 30387.1 MiB
ggml_metal_synchronize: error: command buffer 0 failed with status 5
error: Insufficient Memory (00000008:kIOGPUCommandBufferCallbackErrorOutOfMemory)
ggml_metal_graph_compute: backend is in error state from a previous command buffer failure
```

30.4 GiB against the M4's 22.9 GB working set is a hard OOM, where the 5070 paged to system RAM and
merely crawled. Same missing checkpointing, two different expressions: slow on CUDA, silently wrong on
Metal. After this branch the same run allocates **1387.2 MiB** and trains correctly, so the fix covers
both.

Worth noting separately, because this branch does not close it: the failure was **silent**. The command
buffer failed, ggml logged that the backend was in an error state, and `sa3-train` still exited 0, wrote
an 88 MB `adapter-final.gguf` plus a trainer state, and recorded `loss: 0`. These families no longer
overflow at this size, but the path that converts an OOM into a plausible-looking untrained adapter is
still there for any future one — a larger model, more frames, a smaller Mac. Same class of gap as never
asserting `loss != 0`.

### Timings

Mean of steps 2-6, 128 frames, medium, rank 16:

| family | CUDA after | **Metal (M4)** | Metal/CUDA | allocator |
|---|---:|---:|---:|---:|
| `lora` | 0.48 | 3.04 | 6.3x | 107.3 MiB |
| **`dora-rows`** | 3.30 | **3.61** | **1.1x** | 140.6 MiB |
| `lora-xs` | 0.71 | 7.62 | 10.7x | 435.8 MiB |
| `dora-rows-xs` | 1.24 | 9.73 | 7.8x | 888.9 MiB |
| `dora-cols-xs` | 1.42 | 16.42 | 11.6x | 937.1 MiB |
| `dora-cols` | 1.61 | 18.43 | 11.4x | 937.2 MiB |
| `bora` | 2.17 | 21.29 | 9.8x | 1387.2 MiB |
| `bora-xs` | 1.97 | 18.14 | 9.2x | 1387.1 MiB |

### On "dora-rows is now the slowest of the eight"

That inverts on Metal. Here `dora-rows` is the *second fastest*, and it is the only family that is not
6-12x slower than CUDA — 1.1x against everyone else's 7.8-11.6x.

The split falls exactly along functional vs materialized. `dora-rows` never materializes `W_eff`; the
other seven rebuild it per segment. CUDA has the bandwidth to make that close to free, so the functional
residual matmuls read as pure overhead. The M4 does not — its strength is the F16 simdgroup-matrix path
and it pays heavily for the extra copy traffic. Same asymmetry as #24, where quantization was compute-flat
on Metal but a 33% win on CUDA.

So the follow-up note reads differently with a second backend in hand: it is not that the functional path
costs more per step, it is that CUDA hides what it saves. Worth measuring on Metal before acting on the
CUDA-only reading.

### Caveats

128 frames only; the 512-frame picture may differ. This M4 also throttles over long benchmark sequences,
though the steps 2-6 means matched an earlier independent 2-step run closely, so I do not think it
affected these.

betweentwomidnights:
## Metal validation handoff

Everything below is verified on the PC (CUDA, RTX 5070 Laptop). Metal is the open question.

**Why the risk is low.** Nothing here is a new kernel. `train_lora_effective_weight` uses
`cast` / `mul_mat` / `add` / `scale` / `rms_norm` / `mul` / `transpose` / `cont`, all of which the
monolithic path already ran on Metal — that is how the M4 measured `bora` at ~8.6 s/step in the
first place. The checkpointed decomposition also already worked on Metal for `lora` / `dora-rows`.
This change only moves *where* those ops get built.

**Why check anyway.** What changed is allocator liveness, and buffer reuse is the one area where
backends genuinely differ. `F` being forward-only is what lets all ~228 effective weights fit,
and that is a gallocr assumption rather than a mathematical one. If it does not hold on Metal the
symptom is a memory blowup — which the M4 has enough unified memory to partially mask, so it may
show up as "slow" rather than "failed".

Also worth stating plainly: `bora`, `dora-cols` and the four `-xs` families have **never** run
checkpointed on any backend before this branch.

### 1. Build and test

```
git fetch origin && git checkout fix/dora-cols-training
ctest --test-dir build -C Release
```

Expect 17/17, including the new `sa3-lora-compose-test`.

### 2. Parity — and this is the part that matters

**Compare Metal against Metal, never against the CUDA numbers in this PR.** Different backend
means different arithmetic, so a cross-backend comparison will not match to the last digit and
would read as a regression when it is not. `--checkpoint-backward false` gives the monolithic
reference on the same machine:

```
sa3-train --model medium --dataset <ds> --latents-dir <lat> \
  --adapter-type dora-cols --frames 128 --steps 2 --seed 42 \
  --checkpoint-backward false --out probe-mono

sa3-train --model medium --dataset <ds> --latents-dir <lat> \
  --adapter-type dora-cols --frames 128 --steps 2 --seed 42 \
  --out probe-ckpt
```

`loss` and `grad_norm` must match to the last digit between those two. Worth repeating for at
least `bora` and one `-xs` family. (`dora-cols` monolithic will be slow — that is the bug being
fixed, not a hang.)

### 3. Timing — average steps 10-40, never 2

The first step carries graph build, and for `dora-rows` a one-off `base_norm_sq` reduction over
every targeted weight: 6.17 s against a 1.25 s steady step on CUDA. A short average is mostly
setup and **reorders the families**. An earlier revision of this PR made exactly that mistake and
claimed the functional path was the slowest of the eight; it is third fastest.

```
sa3-train ... --frames 512 --steps 40 --seed 42 --out ss-<family>
```

Also worth capturing: the `[train] block fwd+bwd gallocr buffer: N MiB` line. On CUDA at 512
frames `dora-cols` went from 24,629 MiB (monolithic) to 1,126 MiB.

### What blocks the merge

- a parity mismatch **within Metal**
- the allocator not dropping the way it does on CUDA
- `dora-cols` still crashing

Anything else — Metal slower in absolute terms, different ratios between families — is a data
point for `docs/TRAINING.md`, not a blocker.

### Reference: CUDA steady-state, 512 frames, medium, rank 16

| family | s/step | first step |
|---|---:|---:|
| `lora` | 0.80 | 1.02 |
| `lora-xs` | 1.05 | 1.33 |
| `dora-rows` | 1.25 | 6.17 |
| `dora-rows-xs` | 1.57 | 2.00 |
| `dora-cols-xs` | 1.84 | 2.31 |
| `dora-cols` | 1.95 | 2.35 |
| `bora-xs` | 2.33 | 2.67 |
| `bora` | 2.47 | 2.88 |

Use these for shape, not for equality.

betweentwomidnights:
## Metal validation — checklist complete, nothing blocks

MacBook Air (`Mac16,13`), base M4, **10-core GPU**, 32 GB, fanless — the weakest M4 there is, worth
stating because the timings below are the floor of the family, not a representative Apple number.
Branch `e13d9e0`.

**Verdict against your three blockers:** no parity mismatch attributable to this branch, the allocator
drops exactly as it does on CUDA, and `dora-cols` no longer crashes. Two bugs did turn up, but both
reproduce on `main` at `ccecf60` and neither is caused by this PR.

### 1. Build and test — pass

38/38 ctest, including the new `sa3-lora-compose-test`. (38 rather than 17 because our Metal build
also runs ggml's own test targets.)

### 2. Parity — pass where a reference exists, and the obstacle is itself a result

`lora`, full scope, monolithic vs checkpointed — **exact on both steps**:

```
mono  loss=0.999763 grad=0.202162   loss=0.649902 grad=0.312501
ckpt  loss=0.999763 grad=0.202162   loss=0.649902 grad=0.312501
```

**`dora-rows` drifts at step 2** — mono `0.649841 / 0.73114` vs ckpt `0.649876 / 0.730656`, about
5e-5 on loss and 7e-4 on grad. Step 1 is exact. Since `lora` is exact, this is not generic
reduction-order noise; it is specific to the dora-rows row-norm path on Metal.

I do not think it blocks, for a concrete reason: **`main`'s checkpointed values are byte-identical to
this branch's** (`0.649876 / 0.730656`). This branch changes nothing about what `dora-rows` produces.
It only makes the discrepancy *observable*, because `--checkpoint-backward false` is a flag this PR
adds — `main` has no monolithic mode at all. Worth a look on CUDA (where you report exact matches),
since that would make it Metal-specific.

**The materialized families have no reachable reference on Metal.** Monolithic allocates 19–30 GiB
against a 22.9 GB working set and hard-OOMs:

```
dora-cols mono  alloc=19294.9 MiB  oom  loss=0 grad=0
bora      mono  alloc=30387.1 MiB  oom  loss=0 grad=0
bora-xs   mono  alloc=30387.3 MiB  oom  loss=0 grad=0
```

Dropping to 32 frames barely moved it (18,191 MiB) — the allocation is **frame-independent**, since
it is materializing `W_eff` for 228 targets, which scales with model size rather than sequence
length. No crop makes it fit.

That is worth folding into the PR body as an argument *for* this change: on Metal these five families
are not merely slow without checkpointing, they are **impossible**. The 5070 degraded gracefully into
sysmem paging at 24.6 GiB; the M4 cannot run them at all. And because the OOM leaves the backend in
an error state while `sa3-train` still exits 0 and writes an adapter, the pre-branch failure mode here
was a silent, plausible-looking, completely untrained artifact.

Narrowing scope so a reference fits gave **exact step-1 matches** for all three (`dora-cols`
`0.999763/0.0590135`, `bora` `0.999755/0.0808541`, `bora-xs` `0.999755/0.0854105`) before hitting
bug 1 below.

### 3. Timing and allocator — pass, and the allocator result is the strong one

512 frames, medium, rank 16, mean of steps 10–15.

| family | CUDA s/step | **Metal s/step** | ratio | CUDA first | Metal first | **Metal alloc** |
|---|---:|---:|---:|---:|---:|---:|
| `lora` | 0.80 | 7.32 | 9.2x | 1.02 | 7.23 | 338.8 MiB |
| `dora-rows` | 1.25 | 8.07 | **6.5x** | 6.17 | 9.39 | 407.4 MiB |
| `lora-xs` | 1.05 | 14.33 | 13.6x | 1.33 | 12.49 | 658.3 MiB |
| `dora-rows-xs` | 1.57 | 18.26 | 11.6x | 2.00 | 14.27 | 1112.8 MiB |
| `dora-cols-xs` | 1.84 | 35.46 | 19.3x | 2.31 | 24.30 | 1126.2 MiB |
| `dora-cols` | 1.95 | 37.10 | 19.0x | 2.35 | 24.88 | **1126.3 MiB** |
| `bora` | 2.47 | 40.69 | 16.5x | 2.88 | 26.97 | **1576.3 MiB** |
| `bora-xs` | 2.33 | 40.67 | 17.5x | 2.67 | 26.73 | 1576.2 MiB |

**The allocator lands on your CUDA numbers to the decimal.** You measured `dora-cols` at 1,126 MiB and
`bora` at 1,576 MiB; Metal gives 1,126.3 and 1,576.3. That was the stated risk — `F` being
forward-only is a gallocr assumption, not a mathematical one — and it holds identically here.

Family ordering matches CUDA except that `dora-rows` and `lora-xs` swap. `dora-rows` also has the
lowest Metal/CUDA ratio of the eight (6.5x vs 9–19x), i.e. the functional path is relatively
*favoured* on Metal — consistent with what #24 found, where quantization was compute-flat on Metal
but a 33% win on CUDA. Metal punishes the `W_eff` materialization the other seven families do.

### Deviations from the protocol, and a caveat on two rows

**15 steps, not 40** — at 9–19x CUDA's per-step cost, 40 steps x 8 families is 4+ hours here.

**120 s cooldown before each family.** This is a fanless Air and it throttles hard. Per-step traces
show the heavy families ramping ~50% from cold and plateauing around step 5–7:

```
bora        26.97 27.12 29.55 35.55 39.18 40.33 40.80 ... 40.16
dora-cols   24.88 24.88 26.04 29.80 34.26 36.68 36.55 ... 36.35
lora         7.23  7.24  7.27  7.28  7.29  7.31  7.30 ...  7.33
```

So for the six heavy families, steps 10–15 sit on the throttled plateau and the numbers above are
genuine sustained figures. **`lora-xs` and `dora-rows-xs` had not plateaued by step 15** (still
climbing: 12.49 → 14.97 and 14.27 → 21.97), so those two rows *understate* their sustained cost.
Everything else is solid.

Note this inverts the CUDA convention: there the first step is the slowest (graph build); here later
steps are, because of heat. `dora-rows`' one-off `base_norm_sq` reduction is also relatively cheaper
on Metal — 9.39 against an 8.07 steady step, versus your 6.17 against 1.25.

---

## Two pre-existing bugs — both reproduce on `main`, neither caused by this PR

**1. `--lora-include` segfaults after step 1 (rc=139).**

```
sa3-train --model medium --adapter-type lora --lora-include dit.0. --frames 128 --steps 2 ...
```

Not family-specific — `lora` and `dora-rows` crash too. Not the head-skip path — `--lora-scope core`
also skips `H` and works fine. The trigger is leaving *blocks* unadapted: `--lora-include dit.0.`
gives block 0 adapters and blocks 1–23 none. It looks like the same empty-graph class that #27 fixed
for the head graph (`train_ckpt.h`, "must be SKIPPED rather than built empty"), still unfixed for
block graphs. Confirmed identically on `main` at `ccecf60`.

**2. The checked-in training audio is stubs.** Every file in
`train-runs/ratatat-macos-dataset/train/audio/` is 74–81 bytes; the whole directory is 40K for 10
tracks, and ffmpeg fails to open them. Pre-encoding from the repo dataset is therefore impossible —
training requires an external `--latents-dir`. Probably LFS pointers or evicted files.

---

## Proposed `docs/METAL.md` changes (separate from this PR's subject)

Recording these here rather than pushing doc commits.

**Name the machine in §6, §7 and §9.** They all say only "Apple M4, 32 GB". A reader on an M4 Max
would see wildly different numbers. Should read: MacBook Air (`Mac16,13`), base M4, 10-core GPU,
32 GB, fanless.

**§6's MLX column is against the wrong MLX.** It benchmarks our own `sa3-mlx` research
implementation, which at 120 s recorded 92.68 s generation + 18.74 s setup. Stability's official
`optimized/mlx` does the same 120 s in **34.49 s wall including setup** — ~3.2x faster, and §6 itself
shows why: 76.47 s of our 92.68 s was decode. §6 currently reads as sa3.cpp beating MLX at 120 s
(44.51 vs 92.68); against the official implementation we lose (44.51 vs 34.49). Replace the column.

**§6's 12.14 s row is not comparable to any current measurement.** It predates `775f67a`
(`duration_padding`, 2026-06-30 — one day after that benchmark). At 30 s the default now builds a
389-frame canvas and truncates to 324; in June it built 323. Cold, interleaved, best of 3:

| config | canvas | wall |
|---|---:|---:|
| sa3.cpp default | 389 → 324 | 14.48 s |
| sa3.cpp `--duration-padding 0` | 323 | 11.93 s |
| official MLX, `--steps 8` | 323 | 11.25 s |

So **there is no Metal regression since June** — 11.93 s today vs 12.14 s then, on the same canvas.
And like-for-like we are ~6% behind official MLX, not the ~24% the default-vs-default comparison
suggests. The default simply does ~20% more work than MLX for the same requested seconds, by design.
Any user-facing speed comparison needs to say which of those two things it is measuring.

betweentwomidnights:
## Folded in your bug 1 — and it was not the empty-graph class

`--lora-include` reproduces on CUDA, so this is now fixed here in `2801146`. Thanks for catching
it; the diagnosis went somewhere different from where we both expected.

**It is nondeterministic.** A sweep of `--lora-include dit.N.` over all 24 blocks failed on 8 of
them — and re-running moved *which* 8. So it is not position-dependent and not specific to any
adapter family. It fires on roughly a third of runs, with two faces: `ggml_cuda_set_device`
reporting `invalid device ordinal`, or a plain access violation. Your Metal `rc=139` is the third
face of the same thing.

**Isolation.** Monolithic never fails; checkpointed does:

| path | failures |
|---|---|
| `--checkpoint-backward false` | 0 / 8 |
| checkpointed | 2 / 8 |

**The trigger is heterogeneous block graphs, not the include filter.** This is the part that
explains your `--lora-scope core` observation:

| config | block graphs | failures |
|---|---|---|
| `--lora-scope core` | homogeneous | 0 / 6 |
| `--lora-include dit.` | homogeneous | 0 / 6 |
| `--lora-include dit.0.` | heterogeneous | 2 / 6 |
| `--lora-include dit.0.,dit.1.` | heterogeneous | 2 / 6 |

`--lora-include dit.` runs the identical filter code over all 24 blocks and never fails, so the
filter is fine. `core` keeps every block adapted *uniformly*, which is why it looked like the
head-skip path was the difference — it was not.

**Root cause.** All 24 block graphs shared one `gallocr`. That is safe while the graphs are
structurally identical: gallocr plans once and `ggml_gallocr_alloc_graph` is a replay. Adapt only
some blocks and the graphs differ, so every block boundary trips `ggml_gallocr_needs_realloc`,
whose automatic path is `ggml_gallocr_reserve` — which **frees and reallocates the buffer, 24
times per step, under a live CUDA context**.

**Fix.** Group blocks by graph shape, one allocator per group, each reserved once at build. Two
blocks share an allocator only when their graphs are identical, so no allocator ever re-plans.

**Uniform scopes are untouched** — one shape, one allocator, same buffer, same numbers.
`dora-rows` at 128 frames still reports 140.6 MiB (matching your Metal figure exactly) and
`loss 0.693036/0.869469`, `grad 0.2141/0.2223`. `--lora-include dit.0.` now reports 2 shapes and
206.1 MiB, the honest cost of running two different graphs.

**Verification:** 24-block sweep 0/24 (was 8/24); adapter-type matrix over
`lora`/`dora-rows`/`dora-cols`/`bora` 0/8 (was 6/8); repeated trials of both heterogeneous scopes
0/6 each (were 2/6); 17/17 ctest.

Worth re-checking on Metal, since a freed-and-reallocated buffer under a live command queue is
exactly the kind of thing that could present differently there.

## On your bug 2 — the repo is clean

`train-runs/` has been gitignored since well before this branch, nothing under it is tracked on
any branch, and nothing matching those paths was ever added in history. Those stub files are a
broken local copy of that dataset in your working tree, not something the repo ships. Worth
re-fetching on that machine, but there is nothing to remove here.

## On the dora-rows parity drift

Agreed it does not block, and we are not chasing it. `--checkpoint-backward false` exists as a
debugging aid; checkpointing is the default and is better on every axis, so a 5e-5 difference on
a path nobody runs is not worth the time. If it ever matters, the note in your comment is the
right starting point.

betweentwomidnights:
## `2801146` verified on Metal — segfault fixed. But it exposes a correctness bug underneath.

### The fix works

The `--lora-include` segfault is gone. Deterministic across 3 reps each (it was 100% reproducible on
Metal before, versus the ~30% you saw on CUDA — the shape-mismatched re-plan is evidently more hostile
here):

```
lora_blk0       rc=0 steps=2   (x3, identical)
dora-rows_blk0  rc=0 steps=2   (x3, identical)
```

Uniform scopes are unchanged, so the grouping is inert where every block graph is identical, as
intended.

### The correctness bug it uncovers

With the crash out of the way, step 2 is finally reachable at a narrow scope — and it is wrong.
**Whenever the head graph is skipped, the checkpointed backward diverges after one optimizer step.**
The correlation is 7 for 7:

| run | H graph | step-2 loss |
|---|---|---:|
| `lora` full | `H 184 nodes` | 0.649902 |
| `dora-cols` full | `H 386 nodes` | 0.649906 |
| `lora --lora-include dit.` | `H 184 nodes` | 0.649902 |
| `lora --lora-scope core` | **H skipped** | **3.08571** |
| `dora-cols --lora-scope core` | **H skipped** | **3.08704** |
| `lora --lora-include dit.0.` | **H skipped** | **3.08791** |
| `dora-cols --lora-include dit.0.` | **H skipped** | **3.08711** |

Step 1 is exact in every case. It is step 2 — after the optimizer has consumed the gradients — that
blows up 5x. Monolithic at the same narrow scope is correct, which localises it to the checkpointed
decomposition rather than the adapter maths:

```
dora-cols --lora-include dit.0.  mono  0.999763/0.0590135   0.650287/0.087511   <- correct
dora-cols --lora-include dit.0.  ckpt  0.999763/0.0590135   3.08711/3.40089     <- wrong
```

### It is not this PR

`main` at `ccecf60` produces the identical wrong value:

```
MAIN(ccecf60)  lora --lora-scope core   0.999763/0.0599011   3.08571/5.56061
BRANCH(2801146) lora --lora-scope core  0.999763/0.0599011   3.08571/5.56061
```

Byte-identical, so neither #28 nor `2801146` causes it. It traces to the head-skip path added with
`--lora-scope` in #27 — `train_ckpt.h`'s "must be SKIPPED rather than built empty". Skipping H avoids
the empty-graph assert, but something the head segment was contributing to gradient accumulation is
being dropped rather than legitimately having nothing to contribute.

Worth noting what this means for `main` today: **`--lora-scope core` is a documented, shipped flag
that silently produces broken training after step 1.** Losses are finite and the run completes
normally, so nothing surfaces it — the same shape as the zero-loss families, where the failure looked
like a successful run.

I would not treat it as a #28 blocker, since the branch is byte-identical to main here. But it does
mean the `--lora-scope` flag needs fixing before anyone uses it, and it is worth confirming on CUDA —
if it reproduces there, it is a plain logic bug rather than anything Metal-specific.

### Correction to my earlier comment

I described the audio under `train-runs/ratatat-macos-dataset/train/audio/` as "checked-in". That was
wrong — `train-runs/` is gitignored (`.gitignore:35`, zero tracked files), so those stubs are purely
local test data on this machine and not a repo problem at all. Please disregard that item; the only
consequence was that I needed an external `--latents-dir`, which worked fine.

betweentwomidnights:
## The head-skip divergence does not reproduce on CUDA — it looks Metal-specific

Checked directly, same dataset/seed/frames throughout, 4 steps, only scope and backward mode
varying:

| config | H graph | step 1 | step 2 | step 3 | step 4 |
|---|---|---|---|---|---|
| full, ckpt | `H 184 nodes` | 0.693160 | 0.869186 | 0.294414 | 0.823565 |
| full, mono | — | 0.693160 | 0.869186 | 0.294414 | 0.823565 |
| **core, ckpt** | **H SKIPPED** | 0.693160 | **0.869322** | 0.294487 | 0.823701 |
| **core, mono** | — | 0.693160 | **0.869322** | 0.294487 | 0.823701 |
| `--lora-include dit.0.`, ckpt | **H SKIPPED** | 0.693160 | 0.869450 | 0.294400 | 0.823694 |
| `--lora-include dit.0.`, mono | — | 0.693160 | 0.869450 | 0.294400 | 0.823694 |

Checkpointed equals monolithic **bit-for-bit at every head-skipped scope, through four steps** —
past the optimizer update where you see the blow-up. `core` differs from `full` only in the
fourth decimal, which is what adapting 168 targets instead of 228 should do.

Real runs agree. Two completed 2000-step runs on this box:

| run | first-10 loss | last-10 loss | max |
|---|---:|---:|---:|
| `scope-core-r16-retry` | 1.0075 | 0.8567 | 2.0835 |
| `scope-full-r16` | 1.0075 | 0.8546 | 2.0171 |

`core` converged to within 0.002 of `full` and never exceeded 2.08. (Note the ~1.0075 opening
matches your 0.999763, so we are on comparable data — my 4-step table above uses a different
dataset, hence 0.693.)

So `--lora-scope core` is not silently broken on CUDA, and the flag is not shipping broken
training generally. Whatever you are seeing is specific to Metal.

## A sharper direction for the Metal side

Your own numbers contain a strong constraint worth exploiting:

```
dora-cols --lora-include dit.0.  mono  0.999763/0.0590135   0.650287/0.087511
dora-cols --lora-include dit.0.  ckpt  0.999763/0.0590135   3.08711/3.40089
```

**Step-1 loss and grad_norm are identical.** The optimizer is entirely host-side — it consumes
those gradients and writes new adapter values into `TrainLoraState`. Identical gradients in means
identical weights out, in both modes. So by step 2 the two paths should be applying the *same*
adapter, and the forward should agree.

That points away from the gradient path and toward what step 2 actually reads. The checkpointed
path keeps adapter tensors in one persistent backend buffer and re-uploads them per step
(`run_train_dit_accumulate_ckpt`, the `ggml_backend_tensor_set` loop); the monolithic path uploads
into graph-owned input tensors. If Metal is not making those persistent-buffer writes visible to
the next command buffer, step 2 computes against stale or partially-updated weights — finite,
plausible-looking, and wrong, exactly as observed.

Concrete check: at the top of step 2, read one adapter tensor back with
`ggml_backend_tensor_get` in both modes and compare against the host-side `TrainLoraState` values.
If the checkpointed read disagrees with the host state, it is an upload-visibility problem, not a
gradient problem — and the head-skip correlation is incidental (narrow scopes simply have fewer
targets and reach the divergence sooner or more visibly).

Also worth ruling in or out: does it still diverge at a narrow scope that does **not** skip H
(e.g. `--lora-include dit.proj_in`, head-only)? If that diverges too, H is a red herring
outright. That is the single most informative next run.

## Agreed on the rest

Thanks for the correction on the checked-in audio — matches what I found here (zero tracked files
under `train-runs/`, nothing in history).

And agreed this is not a #28 blocker: byte-identical to `main` on your machine, and clean on CUDA
here.

betweentwomidnights:
## Isolated: the trigger is the H graph being skipped, and it is not the gradient path

Thanks for the CUDA table — that plus your "identical step-1 gradients" constraint pointed me the
right way, though the answer landed somewhere neither of us guessed.

### It is not the optimizer, and not stale adapter uploads

Running with the adapter effectively frozen (`--learning-rate 1e-15`, so AdamW's bounded update is
~0) reproduces the divergence unchanged:

```
core  frozen  mono  0.999763  0.774843  1.99736
core  frozen  ckpt  0.999763  3.08754   2.94811
full  frozen  mono  0.999763  0.649901  1.87495
full  frozen  ckpt  0.999763  0.649901  1.87495
```

If the weights never meaningfully change, the optimizer cannot be the cause — and neither can an
upload-visibility problem on the adapter buffer, since full scope re-uploads through exactly the same
persistent-buffer path and stays bit-exact. Step-1 `grad_norm` is also identical between mono and
ckpt in every pair I ran, so the whole of step 1, gradients included, is correct.

### The trigger is precisely whether `hgraph` exists

Comparing mono vs ckpt **within** each scope (cross-scope comparison is confounded — monolithic
losses legitimately differ between scopes, which is what makes `full` vs `core` an invalid pairing):

| scope | targets | H graph | step-1 grad mono/ckpt | step-2 mono | step-2 ckpt | |
|---|---:|---|---|---:|---:|---|
| full | 228 | `H 184 nodes` | 0.202162 / 0.202162 | 0.649901 | 0.649901 | match |
| head-only (`proj_in,pre_conv,time_embed`) | 4 | `H 98 nodes` | — | 0.649901 | 0.649901 | match |
| `to_qkv,to_out,ff,proj_in` | 49 | `H 47 nodes` | 0.0980365 / 0.0980365 | 0.649901 | 0.649901 | match |
| `--lora-scope core` | 168 | **skipped** | 0.0599011 / 0.0599011 | 0.774843 | **3.08754** | **diverge** |
| `to_qkv,to_out,ff` | 48 | **skipped** | 0.0372422 / 0.0372422 | 0.661931 | **3.08754** | **diverge** |

Every scope that builds H is bit-exact, from 4 targets to 228. Both scopes that skip H diverge.
**Adding a single head target — 48 to 49 — flips it from broken to exact.** So it is not target
count, not partial adaptation, not adapter family, and not `--lora-include` vs `--lora-scope`.

### The strongest hint

The two H-skipped runs produce the **identical** wrong value `3.08754` while adapting completely
different target sets (48 vs 168), where their monolithic counterparts correctly differ (0.661931 vs
0.774843). Whatever the checkpointed path computes at step 2 with H skipped does not depend on which
adapters exist.

Combined with step 1 being perfect, that reads as state carried between steps rather than anything
computed within one. Note the reported loss is read straight after T
(`ggml_backend_tensor_get(ck.loss, ...)`, before the block loop and before H), so H cannot affect the
loss of its own step — only the following one.

One structural difference worth checking: when H is built it also allocates `out.halloc`, and its
compute is the last GPU work in the step. When skipped, the step's final operations are the two
`ggml_backend_tensor_set` calls into `ck.Gctx_in` / `ck.Ggcond_in` — writes whose consumer no longer
exists, with no compute or sync behind them. That is the kind of thing that behaves differently on
Metal's command-buffer model than on CUDA's stream model, and it would explain why CUDA is clean.

### Workaround, and scope of impact

Including any head weight in the scope forces H and restores exact parity — so `--lora-scope core`
plus e.g. `proj_in` is safe today on Metal.

Impact is `--lora-scope core` and any `--lora-include` that misses all 12 head targets, on Metal only,
from step 2 onward. Losses stay finite and the run completes normally, so nothing surfaces it.
Reproduces identically on `main` at `ccecf60`, so it is #27-era and not caused by #28 or `2801146`.

I have not instrumented tensor read-backs yet — say the word if you want me to, though it needs a
code change and you may prefer to do that on the machine you will merge from.

betweentwomidnights:
## Correction: the head-skip divergence is NOT Metal-specific — the CPU backend reproduces it bit-for-bit

I told you this looked Metal-specific. That was wrong, and here is the measurement that overturns it.
Same machine, same scope (`--lora-include to_qkv,to_out,ff`, H skipped), adapter frozen at
`--learning-rate 1e-15`, 3 steps:

| build | mono | ckpt |
|---|---|---|
| **CPU backend** | 0.999763 / 0.661931 / 1.93884 | 0.999763 / **3.08754** / 2.94811 |
| **Metal backend** | 0.999763 / 0.661931 / 1.93884 | 0.999763 / **3.08754** / 2.94811 |
| CPU, H present (`+proj_in`) | 0.999763 / 0.649901 / 1.87495 | 0.999763 / 0.649901 / 1.87495 |

The pure CPU backend on this Mac produces the identical wrong value, to every digit.

**Why that matters more than the divergence itself:** memory corruption, buffer aliasing and
command-buffer races cannot reproduce bit-identically across two completely different backends. This
is deterministic shared logic, not a hardware or driver problem — so it is almost certainly *not*
something to fix in the ggml fork's Metal code, and quite possibly not in ggml at all.

It also means the CUDA result may be platform luck rather than proof of correctness. Your table and
your two 2000-step runs are real evidence that CUDA is fine *there*, but the mechanism is now known to
live in backend-independent code, so I would not assume `--lora-scope core` is structurally safe on
CUDA — only that it did not trigger under the conditions you tested.

### Ruled out, each with a measurement

| hypothesis | disproved by |
|---|---|
| Optimizer / gradient path | frozen adapter (`lr=1e-15`) reproduces it; step-1 `grad_norm` identical mono vs ckpt |
| Adapter upload visibility | step-2 uploads byte-identical (`x`, `target`, `cross`, `global`, `tfeat`, `local`, `lora_A`, `lora_B`) |
| Async read race | explicit `ggml_backend_synchronize` before every probe — no change |
| Scratch aliasing the persistent buffer | measured address ranges for F/T/H and all block graphs — all clear |
| Inpainting (`local` / `loss_weight`) | reproduces with `--inpainting false` (3.19284 vs mono 1.32319) |
| Stale F allocation plan | re-planning F with `ggml_gallocr_alloc_graph` every step — no change |
| Missing end-of-step sync | unconditional `ggml_backend_synchronize` at step end — no change |
| Metal graph-optimize / concurrency / fusion | all three disable knobs — no change |
| `pos` / `ones` (uploaded once at build) corrupted by backward | checksummed each step — stable and identical |

### What is established

- Trigger is exactly whether `hgraph` is built. 7/7 across scopes from 4 to 228 targets; adding one
  head target (48 → 49) flips it from broken to bit-exact.
- Step 1 is fully correct — loss, gradients, and every persistent tensor at end of step.
- At step 2, F has byte-identical inputs and produces different `xb[]`, while `context_p` and
  `gcond_p` from the same `dit_head` call are identical.
- The two F graphs are structurally identical apart from 6 inserted nodes (`proj_in`'s LoRA residual).
- `install_overrides` returns empty for functional families, so F is purely `dl`-driven — the forward
  is the same logic in both.

### A methodology warning

My per-node checksum dump was invalid and I want that on the record: gallocr reuses scratch buffers,
so reading node outputs after the graph completes returns whatever last occupied that memory. It
reported 3048 "divergences" including a dramatic one at node 0, essentially all artefacts. Only the
structural alignment from that pass is trustworthy.

### Workaround

Include any head weight in the scope and H is built again, restoring exact parity — `proj_in` alone is
enough. So `--lora-include to_qkv,to_out,ff,proj_in` is safe today, and full scope (228, the default)
is unaffected and verified bit-exact through 6 steps.

Impact is `--lora-scope core` and any `--lora-include` matching none of the 12 head targets. Losses
stay finite and runs complete normally, so nothing surfaces it. Reproduces on `main` at `ccecf60`, so
it is #27-era and not caused by #28 or `2801146`.

betweentwomidnights:
## The CPU backend does not reproduce it here — and there is an inconsistency inside the M4 data worth resolving first

Ran your exact conditions on this box: your include string, `--learning-rate 1e-15`, 3 steps,
medium, seed 42, monolithic vs checkpointed. Scope resolves identically to yours — 48 targets
(48 per-block, 0 elsewhere, H skipped) and 49 with `proj_in` (1 elsewhere, H built), so we are
running the same thing.

| backend | config | step 1 | step 2 | step 3 |
|---|---|---:|---:|---:|
| CPU | `ff`, mono | 0.693177 | 0.869466 | 0.294438 |
| CPU | `ff`, **ckpt (H skipped)** | 0.693177 | 0.869466 | 0.294438 |
| CPU | `ff+proj_in`, mono | 0.693177 | 0.869466 | 0.294438 |
| CPU | `ff+proj_in`, ckpt | 0.693177 | 0.869466 | 0.294438 |
| CUDA | all four of the above | 0.693160 | 0.869359 | 0.294361 |

All four are bit-identical within a backend. The pure CPU build here does not diverge.

### The part that matters more than the disagreement

At `--learning-rate 1e-15` the adapter cannot move, and `lora_B` is **zero at init**
(`train_lora.h`, "Reference init … and B = 0"), so `delta = B@A = 0`. The adapter is a no-op at
step 1 and stays one. Adding `y + 0` is exact in floating point, so for plain `lora`:

> **with a frozen adapter, the loss trajectory cannot depend on the scope at all.**

That is the invariant my table satisfies — all four configs identical, across two backends.

Your table does not satisfy it, and the violation is on the **monolithic** side, where
checkpointing is not involved:

```
CPU  ff            mono   0.999763 / 0.661931 / 1.93884
CPU  ff+proj_in    mono   0.999763 / 0.649901 / 1.87495
```

Two monolithic runs, frozen adapter, same seed and data, differing only in whether one head
target is in scope — and they disagree at steps 2 and 3. Nothing in the maths can produce that.
So something is varying between your runs beyond the flag under test, which would also explain a
spurious mono-vs-ckpt gap.

I would resolve that before treating the head-skip divergence as real. Two cheap checks:

1. **Run the same command twice, unchanged.** If the two runs disagree, the environment is
   nondeterministic and every comparison in the last few rounds is suspect.
2. **Two monolithic runs, scopes `ff` and `ff+proj_in`, `lr=1e-15`.** These must be identical. If
   they are not, the problem is upstream of checkpointing entirely.

A stale binary is worth ruling out too — a `main`-built `sa3-train` on `PATH` while testing the
branch would produce exactly this kind of stubborn, self-consistent-looking result.

### On the generalisation

> I would not assume `--lora-scope core` is structurally safe on CUDA — only that it did not
> trigger under the conditions you tested.

Fair as stated, but the conditions now include: both backends available here, four scopes, mono
vs ckpt, through 4 steps at normal LR and 3 steps frozen, plus two completed 2000-step runs where
`core` landed within 0.002 of `full`. Combined with the invariant above, I do not think there is a
latent structural issue on this side.

### On the workaround

Holding off documenting "add `proj_in` to your scope" until the inconsistency above is resolved.
If it turns out to be a measurement artefact, documenting the workaround would tell users to work
around a bug that does not exist, and `--lora-scope core` is a shipped flag — the cost of
wrongly disclaiming it is real.

Genuinely useful call on the checksum-dump methodology warning, by the way. gallocr reusing
scratch means post-hoc node reads are meaningless, and that is exactly the trap this codebase
keeps setting — same family as the 2-step timing average and the monolithic OOM that still
exits 0.

betweentwomidnights:
## M4 `--lora-scope core` 2000-step run — full logs for review

Full artifacts (metrics.jsonl for all 2000 steps, config snapshot, command, derived summary):
**https://gist.github.com/betweentwomidnights/f79f7e27b039a1630ea83bfc4b232697**

Ran the end-to-end test rather than more 3-step comparisons, since a real adapter answers the
practical question directly. **It converged, but the adapter does not carry the style**, and kev's
read on the outputs is that something is genuinely wrong — which matters because he reports that on
the PC, 168 targets vs 228 costs almost nothing in strength or quality when healthy.

### Setup

MacBook Air (`Mac16,13`), base M4, 10-core GPU, fanless. Metal backend. Branch at `2801146` (+ docs
`e13d9e0`), **clean source — my instrumentation was stashed and both builds rebuilt before this run**.
Recipe matches the reference: dora-rows rank/alpha 16, `--lora-scope core` (168 targets, H skipped),
512 frames (~23.8 s random crops), seed 42, 2000 steps, the 10-track ratatat pre-encoded latents.

### Loss curve

| block | steps | mean_loss | mean_gnorm | mean s/step |
|---|---|---|---|---|
| 0 | 1-100 | 1.9516 | 3.8455 | 10.85 |
| 1 | 101-200 | 1.1173 | 3.9888 | 13.44 |
| 2 | 201-300 | 0.7420 | 0.6488 | 12.93 |
| 3 | 301-400 | 0.6804 | 0.3083 | 16.31 |
| 4 | 401-500 | 0.6446 | 0.2278 | 12.13 |
| 6 | 601-700 | 0.6219 | 0.2193 | 12.38 |
| 19 | 1901-2000 | 0.6332 | 0.1785 | 15.47 |

`first-10 = 1.9790`, `last-10 = 0.6018`, `mean = 0.7332`, `max = 5.8229`.

### The two things I would most like compared against your runs

**1. It plateaus at ~step 400 and never improves again.** Blocks 4 through 19 sit in a 0.62-0.67 band
for 1600 consecutive steps while grad norms decay to ~0.18. Does `scope-core-r16-retry` show a curve
that keeps descending across the same window, or does it flatten similarly?

**2. Lower final loss than your run, yet weaker style.** We reach `0.6018` where your
`scope-core-r16-retry` reported `0.8567` — but the adapter recalls Ratatat much less. Lower loss with
worse transfer is the part I cannot explain, and it is the strongest hint that the loss is being
minimised against something other than the intended objective.

Also worth noting: `max = 5.8229` against your `2.0835`. Single-step spikes happen with random crops,
but ours is nearly 3x yours.

### Listening

Generated at strengths 1.0 / 2.0 / 3.0, same prompt and seed, plus base-model controls at matched
seeds. The adapter clearly changes the output — it is not inert — but it does not sound like the
training material at any strength.

### Corrections to my earlier comments

- **Caption mix.** I said this run trained on filename captions. Wrong: sampling the stdout showed
  ~60% real captions (`Chiptune, 111 bpm, B minor`, `folk funk, 103 bpm, D minor`) and ~40%
  filenames. Caveat on that number: `metrics.jsonl` carries no prompt field, and I truncated the
  stdout log to its last 40 lines through my own error, so the split is from a 32-prompt sample.
  Indicative only.
- **"core is inherently weaker".** I floated that `core` might transfer style poorly by design since
  it excludes the 12 conditioning/embedding targets (`proj_in`, `cond_embed`, `global_embed`, `gce`,
  …). kev's experience across many PC runs contradicts it, so I withdraw it — it was speculation
  against your data.

### Still outstanding from me

The head-skip framing and the "not Metal-specific" claim both remain unretracted on this PR pending
this investigation. If the curve above looks abnormal against your runs, the micro-benchmarks may
have been detecting something real after the measurement problems are stripped out; if it looks
normal, I owe a fuller retraction.

betweentwomidnights:
## Ran the matched 2000-step `core` run on CUDA. The gap is real, it is not sampling, and the BLAS lead has concrete support

First, two corrections to how the comparison was being framed — both mine as much as yours.

**1. The earlier PC `core` runs were not the same experiment.** `scope-core-r16-retry` recorded
`frames=512` *and* `duration_sec=23`, and `--duration` wins, so it trained on **247-frame crops**.
Yours trained on **512**. Confirmed from `n_gen + n_ctx` in both metrics files: 247 constant vs 512
constant. So `0.8567` vs `0.6018` was never comparable. (Also: 512 frames is ~47.6 s, not 23.8 s —
256 is 23.8.) I have now run a properly matched one: CUDA, `--lora-scope core`, `--frames 512`,
seed 42, same `ratatat-3-1784005242` latents, dora-rows r16.

**2. Seed 42 does not give us the same samples.** Our dataset directories differ, so file
enumeration differs and the RNG stream diverges:

| step | M4 track / mask | PC track / mask |
|---|---|---|
| 1 | Montanita / segments | Wildcat / segments |
| 2 | Loud Pipes / full | Montanita / full |
| 3 | Tacobel / full | Gettysburg / segments |

Any step-by-step loss comparison between our machines is therefore meaningless, including the
early-block numbers I was about to quote at you.

### Normalising by mask type, the difference survives

Grouping the first 300 steps of each run by mask, where the sampling statistics come out nearly
identical (mask counts 62/177/61 vs 69/171/60; mean `n_gen` 277/512/199 vs 234/512/204):

| mask | M4 mean loss | PC mean loss | M4 mean gnorm | PC mean gnorm |
|---|---:|---:|---:|---:|
| `full` | 1.2861 | 0.6983 | **2.7585** | **0.0470** |
| `segments` | 1.2401 | 0.5877 | **2.6708** | **0.0569** |
| `causal` | 1.2551 | 0.6329 | **3.1796** | **0.0724** |

**~2x the loss and ~50x the gradient norm**, consistent across all three mask types. With
`grad_clip` at 1.0, every early M4 step is clipped while the PC never approaches it — so the
adapter is driven far harder than the recipe intends, which fits "converges lower, transfers
worse, sounds perturbed".

Note also that your loss is nearly flat across mask types (1.24-1.29) while the PC's tracks the
generate size (0.59 at ~200 frames, 0.70 at 512). That is its own signal.

### The BLAS lead has a concrete basis

This is a real build-level asymmetry, not a guess:

| | `GGML_BLAS` default | `GGML_ACCELERATE` |
|---|---|---|
| Apple | **ON**, vendor `Apple` (`ggml/CMakeLists.txt:97-98`) | ON -> compiles `GGML_USE_ACCELERATE` into ggml-cpu (`ggml/src/ggml-cpu/CMakeLists.txt:60-69`) |
| Windows / Linux | **OFF** (`ggml/CMakeLists.txt:101-102`) | inert, guarded by `if (APPLE ...)` |

So your CPU **and** Metal builds both link Accelerate into the CPU backend and register a BLAS
accel device; none of my builds do. That is exactly the shape of the evidence: your two backends
agree bit-for-bit with each other and differ from every backend here.

### The test that settles it, entirely on your machine

Do not compare against my numbers — compare against yourself, which removes every confound
(dataset, captions, file order, RNG):

```
cmake -B build-metal-noblas -DGGML_METAL=ON -DGGML_BLAS=OFF -DGGML_ACCELERATE=OFF
cmake --build build-metal-noblas --config Release -j
```

Then run 300 steps of the identical command on both builds and compare **mean gnorm grouped by
mask**. If the no-BLAS build drops to ~0.05, BLAS is confirmed and this stops being a
`--lora-scope` question entirely. If it stays at ~2.7, BLAS is exonerated and the head-skip thread
is live again.

Worth doing the same 300 steps at **full scope** on both builds too. If full scope also shows
~2.7 gnorm, then this was never about `core` — it affects all Apple training, and the head-skip
correlation was coincidence.

### Remaining confounds, stated plainly

Our runs still differ in dataset directory and prompt config — your captions are ~60/40
real/filename, mine are 100% real since `use_paths` went off. Captions plausibly move the loss.
They do not plausibly move the gradient norm 50x, which is why I am treating gnorm as the signal
and loss as supporting. The with/without-BLAS comparison on your machine has neither confound.

betweentwomidnights:
## Matched 2000-step `core` run finished on CUDA — two of your three observations dissolve, one sharpens

Same config as yours: `--lora-scope core`, `--frames 512`, seed 42, dora-rows r16, same
`ratatat-3-1784005242` latents, 2000 steps. 1938 s total (0.969 s/step), allocator 395.1 MiB,
1 graph shape.

### 1. The plateau is normal

You asked whether our curve keeps descending across steps 400-2000 or flattens like yours. **It
flattens.**

| block | M4 loss / gnorm | PC loss / gnorm |
|---|---|---|
| 0 | 1.9516 / 3.8455 | 0.6734 / 0.0481 |
| 2 | 0.7420 / 0.6488 | 0.6629 / 0.0637 |
| 4 | 0.6446 / 0.2278 | 0.6800 / 0.0862 |
| 6 | 0.6219 / 0.2193 | 0.6251 / 0.0923 |
| 9 | 0.6288 / 0.2056 | 0.6165 / 0.1116 |
| 14 | 0.6365 / 0.1706 | 0.6026 / 0.1227 |
| 19 | 0.6332 / 0.1785 | 0.6086 / 0.1359 |

Both sit in a 0.60-0.68 band from block 4 onward. Nothing abnormal about a 1600-step plateau here.

### 2. "Lower final loss yet weaker style" was the crop mismatch — and it inverts

You reached `0.6018` against what you thought was our `0.8567`, and could not explain lower loss
with worse transfer. That `0.8567` came from a **247-frame** run. Properly matched at 512 frames,
we land at **`0.5749`** — *lower* than yours, and on kev's ears the style transfer here is fine.

So the paradox is gone: lower loss goes with better style, as you would hope. There is no
"minimising the wrong objective" effect to explain.

Your `max = 5.8229` vs our `2.0835` was the same artifact. Matched, ours is `1.0425`.

### 3. What actually remains: gradient magnitude

Full-run means grouped by mask, with the mask distributions nearly identical (415/1169/416 vs
405/1196/399):

| mask | M4 loss | PC loss | M4 gnorm | PC gnorm |
|---|---:|---:|---:|---:|
| `causal` | 0.6948 | 0.5903 | **0.6489** | **0.1187** |
| `full` | 0.7698 | 0.6662 | **0.5880** | **0.0978** |
| `segments` | 0.6689 | 0.5489 | **0.5208** | **0.0929** |

Sustained across the whole run: loss ~15-22% higher, **gradient norm 5.5-6.6x higher**. And the
consequence that matters:

> **M4 clips on 179 of 2000 steps (9.0%). CUDA clips on 0 of 2000 (0%).**

With `grad_clip` at 1.0, 9% of your updates have their magnitude overridden — concentrated in the
first ~200 steps, where gnorm sits near 4.0 and every step is clipped. That is a different
optimisation trajectory from the same recipe, and it is a plausible mechanism for an adapter that
converges normally but carries less of the material.

Note the two gnorm curves run in opposite directions: yours starts ~3.9 and decays to 0.18, ours
starts 0.048 and climbs to 0.136. Ours is what I would expect as an adapter gradually acquires
signal from zero-init `B`.

### Where that leaves it

The `--lora-scope core` framing looks dead: the scope behaves normally here at the matched config,
and your own run converges to a normal place. What is left is a machine-level difference in
gradient magnitude, present from step 1, which is exactly what the with/without-BLAS comparison on
your machine will confirm or kill. The clip rate is a good single number to compare — it needs no
cross-machine calibration.

### One correction that is ours, not yours

Your "512 frames (~23.8 s)" came from **our** docs. `TRAINING_BENCHMARKS.md` stated the headline
recipe as "512 latent frames (~23.8 seconds)" and listed 256 as "~11.9 s" — an off-by-2x from
treating a latent frame as 2048 samples instead of 4096. Fixed on main in `725c4ba`, verified by
measurement (`sa3-generate` reports 23.78 s at 256 and 47.55 s at 512, on both medium and
small-music). So the crop-length confusion was a documentation bug on our side, not a
misreading on yours.

betweentwomidnights:
## BLAS is exonerated. The scope gap is real, and it is scope-specific.

Ran the prescribed test — same machine against itself, 300 steps, four runs, identical command,
only the build and `--lora-scope` varying. That removes dataset, captions, file order and RNG as
confounds.

| run | mean loss | mean gnorm | causal | full | segments |
|---|---:|---:|---:|---:|---:|
| `core_BLAS` | 1.2312 | 2.6253 | 2.4828 | 2.6604 | 2.6680 |
| `core_noBLAS` | 1.2312 | **2.6253** | 2.4828 | 2.6604 | 2.6680 |
| `full_BLAS` | 0.7629 | 0.6192 | 0.6185 | 0.7053 | 0.3703 |
| `full_noBLAS` | 0.7629 | **0.6192** | 0.6185 | 0.7053 | 0.3703 |

### BLAS: out

Bit-identical to four decimals across all three mask types. Linkage verified rather than assumed:

```
build-metal/bin/sa3-train        -> libggml-blas.0.dylib   (present)
build-metal-noblas/bin/sa3-train -> (no libggml-blas)
```

Both load 0.16.0. So the BLAS backend really was linked in one and absent in the other, and it
made no difference whatsoever. In hindsight that is unsurprising: with Metal enabled the scheduler
places this work on Metal, so the BLAS accel device is registered but never selected. The
Apple-vs-Windows build asymmetry is real, it just is not the mechanism here.

### Both of your decision branches point the same way

> If it stays at ~2.7, BLAS is exonerated and the head-skip thread is live again.

`core` stayed at **2.6253**.

> If full scope also shows ~2.7 gnorm, then this was never about `core`.

`full` came in at **0.6192**, not ~2.7. So it is scope-specific, not all Apple training.

**`core` carries 4.2x the gradient norm and 1.6x the loss of `full` on the same machine.** With
`grad_clip` at 1.0 that means the scoped run is clipped on most steps while full scope rarely is —
which is the mechanism you identified, and it fits the listening result: the 2000-step `core`
adapter perturbs the output audibly but does not carry the style, while kev's full-228 Metal runs
produce good Ratatat.

This also retroactively supports the head-skip correlation. Whatever measurement problems I
introduced around it — and there were several — the 48→49-target flip survived all of them.

### Open question for the PC

Our Metal full-scope gnorm is **0.6192** against your core-scope ~0.05. Different scopes, so not
comparable. **What is your full-scope 300-step gnorm?** If it is also ~0.05, then Apple is elevated
at *both* scopes and there are two effects stacked rather than one.

### On #28

Nothing here blocks it. The scope behaviour reproduces on `main` at `ccecf60`, so it is #27-era and
orthogonal to this PR's subject, and #28's own gates passed cleanly on Metal — 38/38 ctest,
`dora-cols` fixed, all eight families training, and the allocator landing on your CUDA numbers to
the decimal (1,126.3 vs 1,126 MiB; 1,576.3 vs 1,576 MiB).

betweentwomidnights:
## PC full-scope 300-step numbers — and the answer is "both, and the ordering inverts"

Fresh 300-step run, matched to my `core` run in everything but scope (medium, `--frames 512`,
seed 42, dora-rows r16, same latents, same captions — so no confound between the two):

| run | mean loss | mean gnorm | causal | full | segments | clipped |
|---|---:|---:|---:|---:|---:|---:|
| PC `core` | 0.6611 | **0.0548** | 0.0724 | 0.0470 | 0.0569 | 0/300 |
| PC `full` | 0.6599 | **0.1529** | 0.1947 | 0.1375 | 0.1486 | 0/300 |

### Answering your question directly: Apple is elevated at both scopes

| scope | Metal gnorm | PC gnorm | ratio |
|---|---:|---:|---:|
| `full` | 0.6192 | 0.1529 | **4.0x** |
| `core` | 2.6253 | 0.0548 | **47.9x** |

So yes — two effects stacked, as you suspected. A ~4x baseline elevation present even at full
scope, plus something far larger that only appears when the scope is narrowed.

### The sharper fact: the scope ordering is inverted between machines

| | `core` -> `full` |
|---|---|
| PC | 0.0548 -> 0.1529 (**full is 2.8x higher**) |
| Metal | 2.6253 -> 0.6192 (**core is 4.2x higher**) |

On the PC, widening the scope *raises* the global gradient norm. That is what should happen: it is
an L2 over all trainable vectors, so going from 168 to 228 targets adds terms — and the 12
head/tail targets it adds (`proj_in`/`proj_out`, `pre_conv`/`post_conv`, the embeddings) evidently
carry larger per-parameter gradients than the block projections, since the increase is 2.8x rather
than the sqrt(228/168) = 1.16x you would get from counting alone.

On Metal it goes the other way: **removing** 60 targets *quadruples* the norm. Nothing about
adding or removing terms from an L2 sum can do that. That says the scoped case is producing
inflated gradients rather than the full case suppressing them, which is a much more specific claim
than "core is elevated".

Loss agrees. On the PC, `core` and `full` land on essentially the same loss (0.6611 vs 0.6599) —
they are training the same model on the same data, so they should. On Metal `core` sits at 1.2312
against `full`'s 0.7629, a 1.6x gap that has no counterpart here.

Also note neither PC run clips at all: 0/300 at both scopes, against your 9% at `core`.

### Where I would look next

Given BLAS is out and the effect scales with *which* targets are excluded rather than how many,
the 48->49 flip you kept finding is probably the real signal after all. The head/tail targets are
exactly the ones whose gradients arrive through the `T` and `H` graphs rather than the per-block
`B_l` graphs, and `H` is the one that gets skipped. On the PC those same 12 targets are the ones
that account for full scope's 2.8x higher norm — so they are gradient-heavy on both machines, and
on Metal their absence coincides with the blow-up.

That points at the `T`/`H` contribution to `dL/dcontext` and `dL/dgcond`, which is accumulated
host-side across every block and then either consumed by `H` or, when `H` is skipped, uploaded to
`Gctx_in`/`Ggcond_in` and dropped. Worth checking whether those accumulated vectors differ between
the two scopes on Metal in a way they do not here — a single `ggml_backend_tensor_get` on
`Gctx_in` after the block loop, `core` vs `full`, same step.

Happy to run any specific counter-measurement on CUDA that would help bracket it.

betweentwomidnights:
## Localised: inflation enters through the per-block graphs, and only `H` executing fixes it

Took your suggestion and instrumented the accumulation stage by stage (pre-clip global L2 after
`T`, after the `B_l` loop, after `H`, plus the norms of the host-side `gctx_sum` / `ggcond_sum`).
Metal, 512 frames, seed 42, step 2.

### `core` vs `full`

| stage | `core` | `full` |
|---|---:|---:|
| after `T` | 0.000000 (0 tail params) | 0.826411 (2 tail params) |
| after `B_l` | **5.636396** | 0.941640 |
| `\|gctx_sum\|` | **0.404941** | 0.006271 |
| `\|ggcond_sum\|` | **9.243105** | 1.873588 |
| after `H` | 5.636396 (skipped) | 2.489338 (ran) |

The block graphs contribute ~5.64 in `core` against ~0.45 in `full` — about 12x. And the
accumulated activation gradients you asked about are inflated exactly as you predicted:
`gctx_sum` **64x**, `ggcond_sum` **5x**.

**At step 1 they match**: `gctx_sum` is 0.000457 in both and `ggcond_sum` 0.321407 vs 0.321394. So
nothing is wrong at init — step 1's update creates the divergence.

### Ruling out "a segment with zero trainable params"

`core` has **zero tail params as well as zero head params**, so `T` also runs with nothing
trainable. That looked like the same hazard the `H` guard exists for, so I split it — three runs at
72/73 targets differing by one target:

| run | targets | `T` contributes | after `B_l` | `\|gctx_sum\|` | `\|ggcond_sum\|` | gnorm |
|---|---|---:|---:|---:|---:|---:|
| blocks only | 72 (0 elsewhere) | 0.000000 | 4.982893 | 0.383294 | 8.885983 | **4.98** |
| + `proj_out` (tail) | 73 (1 elsewhere) | 0.615436 | 5.007677 | 0.386406 | 8.841840 | **5.01** |
| + `proj_in` (head) | 73 (1 elsewhere) | 0.000000 | 0.437671 | 0.006269 | 1.872781 | **0.59** |

Giving `T` a real parameter changes nothing — blocks stay at 5.01. Only the run where `H` executes
drops, to 0.59. So it is specifically `H`, not zero-parameter segments in general, and the 48->49
flip was pointing at the right thing.

### What this narrows it to

`H` runs *after* the block loop, so it cannot alter the same step's block gradients. The chain has
to be: `H` skipped -> step 1's update differs -> step 2's activation gradients explode. That is
consistent with step 1 matching and step 2 diverging.

What I cannot explain is the direction. On your CUDA numbers, narrowing the scope *lowers* the norm
(`full` 0.1529 -> `core` 0.0548). On Metal, narrowing it raises the norm 4.2x. Same host-side
optimizer code, opposite sign.

### Counter-measurement that would help

If you can run `--lora-include self.qkv,self.out,ff.proj` (72 targets, H skipped) on CUDA with the
same stage-by-stage breakdown, the single most useful number is **`|gctx_sum|` and `|ggcond_sum|` at
step 2**. On Metal those are 0.383 / 8.886 with H skipped versus 0.006 / 1.873 with H present. If
CUDA shows the same ~60x gap in `gctx_sum` but without the blow-up in the block gradients, the
activation-gradient accumulation is a red herring and the fault is in how the step-1 update is
applied. If CUDA shows no gap at all, the accumulation itself diverges on Metal and that is where to
dig.

Instrumentation is a ~20-line env-gated block (`SA3_GRAD_DEBUG=1`) on
`experiment/metal-lora-scope-gnorm`, branched off `79f57b8`. Happy to push it if you want to run the
identical probe.

PR #30, opened 2026-08-14
quant: fetch quantized encodings, and train LoRAs on a quantized base

betweentwomidnights:
Two things, both prerequisites for publishing quantized GGUFs, plus the release plumbing they turned
out to need.

## 1. The downloaders can fetch quantized encodings

`build_download_plan` hard-rejected anything but F16/F32, so the quants we have been producing since
the q4/q8 work were unreachable through the documented path. `model_artifacts.py`, `models.sh`,
`models.cmd` and `download_models.py` now accept `q4_k_m` / `q5_k_m` / `q8_0` for the DiT and SAME;
conditioner, encoder and tokenizer stay F32.

`download_models.py` was the straggler — it kept `choices=["f16","f32"]` and had no `--dry-run` long
after `model_artifacts.py`'s own comment claimed `--dry-run` printed the resolved plan, and it is the
script every model card prints as the way to fetch models. `--dry-run` now runs before the
`huggingface_hub` import, so the plan is inspectable without the dependency installed.

`docs/DISTRIBUTION.md`'s quant matrix was stale in two ways: it still called integer quants "a later
pass", and it said to keep the SAME at F16/F32 — a precaution written before the tooling existed.
`sa3-quant-check` reports `below-threshold=0` at `0.990` for every DiT and SAME at all three tiers
across all three variants, and `sa3-generate` resolves DiT and SAME with the same encoding suffix,
so shipping a quantized DiT without a matching SAME would break `--encoding` outright.

small-sfx had no quants at all; it now has the same three tiers as the other two variants.

## 2. LoRA training on a quantized base

The forward already worked — the functional adapter path only ever passes the frozen `W` to
`mul_mat`, which is what made quantized *inference* work in #24. The backward is
`out_prod(W, transpose(grad))`, and that was the one op that refused a quantized `src0`.

- **`train_lora.h`** — `train_tensor_to_f32` asked `ggml_backend_tensor_get` for `n` floats from a
  quantized tensor, over-reading it and tripping the bounds assert **before any graph was built**,
  so every family needing base norms aborted regardless of backend. Now delegates to `lora.h`'s
  `read_to_f32`.
- **ggml PR #4** (`ba817aa1`) — quantized `src0` for `OUT_PROD` on CUDA, Vulkan and Metal. Two
  index-convention traps, one per backend family, and **both invisible to a Q8_0 test**; the fork's
  own type coverage reached Q4_K and Q8_0 but not Q5_K or Q6_K, and a `q4_k_m` mix emits Q6_K, so
  the type that ships was untested on all three backends until this PR added cases for it.

### It is faster, not a tradeoff

Steady state, mean of steps 6-15 — a two-step mean is mostly graph build and misled me twice here:

| backend | Q4_K_M | F16 | |
|---|---:|---:|---|
| CUDA | **0.93 s** | 1.08 s | medium, 128 frames, dora-rows |
| Vulkan | **1.72 s** | 1.90 s | same |
| Metal (M4) | **2.95 s** | 3.39 s | counterbalanced ABBA with cooldowns |
| CPU | **29.2 s** | 46.9 s | 64 frames, plain lora |

On a 2.9x smaller base (962 MB vs 2773 MB) and 35% less VRAM (3414 vs 5243 MiB peak).

Incidentally: an **F32** base is 27.72 s/step, 25x slower, because it peaks at ~7600 MiB of an
8151 MiB card. VRAM pressure, not arithmetic — the same signature as #28.

### It costs nothing in quality

2000 steps, medium, 512 frames, seed 42, dora-rows r16, the ratatat set, identical in everything but
base precision. Loss tracks 2.5-4% higher throughout, the offset a slightly degraded forward should
produce, and kev's verdict on the pair that matters is **audibly identical**.

The QLoRA hypothesis proper — that a q4-trained adapter is more stable when deployed on q4 — came
out weakly supported and is *not* claimed here: 0.80 vs 0.76 cosine on one prompt and one seed is
not wide against how much those move with seed.

## 3. Release plumbing

Q4_K_M training bases are now published artifacts, so `TRAINING_BASE_ENCODINGS` makes the F16
fallback a property of *what is published* rather than of what works: `q4_k_m` resolves to a Q4_K_M
base, `q5_k_m`/`q8_0` still fall back. Cards get encoding tables with real sizes, the Metal caveat
(footprint everywhere, speed only on CUDA/Vulkan — Metal is flat, Q4_K_M 2.1% *slower*), and
`base_model_relation: quantized`.

## Validation

- `test-backend-ops -o OUT_PROD`: **Vulkan 128/128** on both devices, **CUDA 224/224**,
  **Metal 128/128**, plus 12884/12884 on the full Metal sweep.
- CUDA and Vulkan agree to the fifth decimal on a real training step.
- F16 training is unchanged and bit-identical to the pre-change reference. 18/18 ctest at the
  merged pin.
- Both new base quants: `below-threshold=0` at `0.990`, and small-music trains on its Q4 base.


betweentwomidnights:
## Trained a real adapter on a quantized base. It is indistinguishable from the f16-trained one.

2000 steps, medium, 512 frames, seed 42, dora-rows r16, the ratatat set. Identical in everything
except the frozen base precision — including the caption set, which is why this uses a **fresh** f16
control rather than the older `ratatat-ab-dora-rows` run (that one trained ~45% on filename captions,
before `use_paths` was turned off; comparing against it would have confounded base precision with
caption quality).

| base | mean s/step | block-0 loss | block-19 loss |
|---|---:|---:|---:|
| Q4_K_M | 1.023 | 0.6983 | 0.6234 |
| F16 | 1.000 | 0.6723 | 0.6075 |

Same curve shape throughout, q4 running 2.5-4% higher — the offset a slightly degraded forward
should produce. Note q4 is marginally *slower* here at 512 frames; the 14% win reported above was at
128 frames, so the speed advantage is not uniform across crop size. The footprint advantage is
(962 MB vs 2773 MB base, 3414 vs 5243 MiB peak VRAM).

### Listening

Both adapters rendered at 8 steps / CFG 1 on a real training caption, across both deployment
precisions. kev's verdict on the pair that matters — `trainedQ4-onQ4` vs `trainedF16-onQ4` — is
**audibly identical**.

So the practical result is that training on a quantized base costs nothing in quality. That is what
makes it interesting for small hardware: train and run inference entirely in q4.

### On the QLoRA hypothesis specifically

Weakly supported, not established. How much each adapter changes when moved between bases:

| adapter | cos(on-q4, on-f16) |
|---|---:|
| trained on q4 | 0.8029 |
| trained on f16 | 0.7619 |
| none (base only) | 0.1652 |

The q4-trained adapter is more stable across bases, which is the predicted direction — but this is
one prompt and one seed, and 0.80 vs 0.76 is not wide against how much these cosines move with seed.

Worth noting the mechanism is weaker here than in the LLM case anyway: the adapter trains against the
quantized **`-base`** DiT but deploys onto the quantized **inference** DiT, which is a different
tensor set (post-trained/ARC). Any benefit comes from training against quantization-shaped weights in
general, not from matching the exact weights.

### What this means for Metal

The feature is worth finishing. A phone or a small Mac could train and run a LoRA entirely in q4, and
the quality question is now answered rather than assumed. `METAL_OUT_PROD_QUANT_HANDOFF.md` on this
branch has the implementation detail, including the two traps: the even-`iqs` pair mapping, and that
**Q8_0 passing proves nothing** because it masks the bug that garbles Q4_K.

Issue #1, opened 2026-07-04
error: [gguf] short read te.embed.weight


hasaranga:
hi,
I get following error:

```
c:\sa3\sa3.cpp>sa3-generate --model small-music --prompt "upbeat funk groove with slap bass" --out song.wav
error: [gguf] short read te.embed.weight
```

I'm using CPU build.

betweentwomidnights:
uh oh! 

could you try this one and see? 

sa3-generate --model small-music --prompt "upbeat funk groove with slap bass" --out song.wav --encoding f16

i think maybe the issue is related to the model loader I just wanted to check if that works first.

hasaranga:
same error!

betweentwomidnights:
very strange. digging into it right now and in a few minutes should have a change pushed so that at the very least the error is more informative... sry and thank you for helping me test this!

hasaranga:
Are you getting the same error for cpu build?

betweentwomidnights:
nope i see this for both types of encoding. 

```
C:\dev\sa3.cpp>sa3-generate --model small-music --prompt "upbeat funk groove with slap bass" --out song.wav --encoding f16
prompt: "upbeat funk groove with slap bass"  (7 tokens, ~11.89s)
  + 6.13s schedule headroom (canvas T=194, truncated back to 128)
  step 1/8  t=1.0000
  step 2/8  t=0.9944
  step 3/8  t=0.9845
  step 4/8  t=0.9579
  step 5/8  t=0.8909
  step 6/8  t=0.7455
  step 7/8  t=0.5125
  step 8/8  t=0.2739
wrote song.wav  (11.89s, seed 0)
```

this is definitely highlighting that i need to make the CLI more helpful though. I had been furiously working on validating the libsa3 in the iPlug2 project and the http server the past day.

hasaranga:
May be damaged model download? It was very slow to download models from huggingface using built-in script. Is there a way to download it using like this tool? https://github.com/bodaay/HuggingFaceModelDownloader

betweentwomidnights:
Thanks, that makes sense. Since both `f16` and `f32` fail at `te.embed.weight`, I think this is probably the shared T5 encoder GGUF being incomplete/corrupt from an interrupted download.

Could you check this file?

```cmd
dir models\t5gemma-b-b-ul2-encoder-0.3B-v1.0-F32.gguf
[...]

PR #34, opened 2026-08-28
Validate shared GGML 0.17 on SA3

betweentwomidnights:
## Purpose

Pin GGML `19c5421c9314893db83ab93575d34b9cb828516f` from
[shared GGML PR #5](https://github.com/betweentwomidnights/ggml/pull/5), the same
candidate used by [ACE PR #1](https://github.com/betweentwomidnights/acestep.cpp/pull/1).
This is a draft hardware-validation PR, not a request to merge before Metal passes.

The only application change is the submodule pin, on top of main `378c269`.
The active `ae-lora` development checkout was not modified.

## Windows validation already completed

Same SA3 source with the old 0.16 dependency vs the new 0.17 candidate:

- Full CUDA and Vulkan Release builds, including CLI, server, and C ABI.
- 18 CTests per revision/backend, all passing.
- 52 model-backed runs per backend: small-music/medium, F16/Q4, functional adapters
  and blends, native preprocessing, checkpointed training/resume, C ABI, adapter reload.
- On each backend, all 15 waveform pairs and all compared adapter/optimizer tensors
  exactly matched the old dependency. No meaningful performance improvement.
- The published GGML commit preserves exactly that tested computational source
  (`d1bead82` plus the latest Metal capability guard); only documentation/history changed.

## M4 checkout and validation

In a clean dedicated SA3 checkout:

```sh
gh pr checkout --repo betweentwomidnights/sa3.cpp feature/shared-ggml-v0.17
git submodule sync --recursive
git submodule update --init --recursive
git -C ggml rev-parse HEAD
# Expected: 19c5421c9314893db83ab93575d34b9cb828516f
cmake -S . -B build-metal -DCMAKE_BUILD_TYPE=Release \
  -DSA3_METAL=ON -DSA3_CUDA=OFF -DSA3_VULKAN=OFF -DBUILD_TESTING=ON
cmake --build build-metal --parallel 4
ctest --test-dir build-metal --output-on-failure
```

Do not use `git submodule update --remote`: test the application's exact pin.
Do not replace an active M4 development checkout; use a separate clone/worktree.

- [ ] Confirm the wide-row test and real model runs select Metal, not CPU fallback.
- [ ] F16/Q4 inference for small-music and medium; functional DoRA and a blend.
- [ ] Native preprocessing and short checkpointed training on both precisions.
- [ ] Compare resumed vs uninterrupted adapter and optimizer tensors.
- [ ] C-ABI training/callbacks and newly trained adapter reload.
- [ ] Report M4 results here, including the tested GGML hash; validate ACE PR #1 too.

Use a separate old-pin checkout for within-Metal numerical comparison. Windows
waveform equality does not establish Metal equality or performance. If GGML fixes
are needed, update shared PR #5 and then bump both application PR pins together.


betweentwomidnights:
## M4 Metal validation — PASS

Apple M4, 32 GB, macOS 15.7.9 (24G830), Xcode 26.2, CMake 4.4.0. Two clean dedicated
clones outside any development checkout; the active `feature/ios-build` tree was not touched.

| | SA3 | GGML |
|---|---|---|
| new | `b1ad295` (main `378c269` + pin) | **`19c5421c9314893db83ab93575d34b9cb828516f`** |
| old | `378c269` | `3d9d897a` |

Configure reported `ggml version: 0.17.0 / commit 19c5421c` and `0.16.0 / 3d9d897a`
respectively. Both `-DSA3_METAL=ON -DSA3_CUDA=OFF -DSA3_VULKAN=OFF -DBUILD_TESTING=ON`,
Release. Both built with **zero compiler warnings**. `ctest`: **18/18 passed on each**.

### Metal was actually used

Every model-backed run logs `[sa3] backend: MTL0 (Apple M4, 21.3 GB)` — 19 inference runs
and 12 training runs, no CPU fallback anywhere in the logs. `sa3-set-wide-row-test`
executes its `GPU(MTL0)` cases explicitly alongside the CPU ones and passes both.
During training the Metal `kernel_out_prod_{f16,f32,q4_K,q6_K}_f32` pipelines are
compiled and dispatched, so the backward pass genuinely runs OUT_PROD on the GPU.

### Waveform equality — 24/24 pairs bit-exact

Same seeds and same local data on both pins; compared as decoded PCM, not just file hashes.

- **Inference (19):** medium F16/Q4_K_M/Q8_0 and small-music F16/Q4_K_M; short and long
  durations; CFG + negative prompt; `--dist-shift Flux`; APG + cfg-rescale.
- **Functional DoRA (6 of the above):** each adapter alone, at strength 0.5, and a
  two-adapter blend (0.6/0.4), on both F16 and Q4 bases.
- **C ABI (2):** `libsa3` generate at pinned seeds 42 and 7.
- **Trained-adapter reload (3):** adapters trained *during this validation* reloaded for
  inference on both pins (F16, Q4, and the C-ABI-trained one).

Sanity checks that keep the equality meaningful: every adapter, strength and blend produces
a *different* waveform from the no-adapter baseline and from each other, so the adapters are
demonstrably applied rather than silently skipped.

### Training, resume and optimizer state — 31/31 comparisons exact

Native pre-encode from raw 44.1 kHz stereo WAV on Metal, `dora-rows` rank 16, 512 frames,
reference seed 42, checkpoint every 3 steps, on both F16 and Q4_K_M bases.

- old vs new, uninterrupted: 10/10 exact (`adapter-step-{3,6}`, `adapter-final`, `trainer-state-step-{3,6}`, both precisions)
- old vs new, resumed: 6/6 exact
- **resumed vs uninterrupted, within each pin: 12/12 exact** — full-state resume reproduces
  the uninterrupted run bit-for-bit on both pins
- adapter conversion (`sa3-lora-convert`, safetensors→GGUF): 2/2 byte-identical
- C-ABI trained adapter: 1/1 byte-identical

All losses and gradient norms finite; adapters change between checkpoints (step-3 ≠ step-6),
so equality is not the degenerate "nothing updated" case.

### C ABI

`sa3-libtrain` completes with `on_step` firing every step (finite loss/gnorm/lr) and a valid
final adapter on both pins. `sa3-libcancel` cancels identically on both (`cancelled
successfully after 4 poll(s)`).

One note for whoever reads the C-ABI results: `tools/sa3-libtest.c` hardcodes `req.seed = -1`,
so back-to-back runs draw different seeds and cannot be compared directly. The two C-ABI
waveform pairs above come from the same tool rebuilt with the seed taken from `argv[4]`.
Not a defect, but worth knowing before someone reads a false mismatch as a regression.

### Performance: no measurable difference

Three timing experiments on `dora-rows`, 512 frames, F16. A naive sequential pass first
suggested the new pin was 8–11% slower, but that was thermal drift, not the pin: in an
alternating A/B/B/A sequence the **old** pin alone degraded from 11.32 to 14.16 s/step as
the machine heated. Thermally settled adjacent pairs agreed to within 0.2%
(old 11.320 / new 11.303, old 11.337 / new 11.318). With 120 s of idle before each run the
scatter between adjacent runs was ±10% in both directions.

Run-to-run variance on this machine exceeds any difference between the pins, so: **no
measurable performance change on Metal**, consistent with the Windows finding. Inference
elapsed times were likewise a wash (-9% to +3%, unsigned noise).

### GGML PR #5 claims, verified against the pinned tree

- `git diff 20fbdd9e 19c5421c` → **only `SA3_PATCHES.md`** (54 added lines). The
  computational source of the pinned commit is identical to the Windows-validated tree.
- `git diff d1bead82 20fbdd9e` → **only** `src/ggml-metal/ggml-metal-device.m`, +6/-1: the
  OUT_PROD `has_simdgroup_mm` capability guard, authorship retained.
- No duplicate CUDA get-rows helpers or GGUF accessors survive in the final tree
  (`ggml_cuda_op_get_rows`, `ggml_cuda_op_get_rows_back`, `gguf_get_tensor_ne` each defined once).

### One gap this hardware cannot close

The OUT_PROD guard is the only Metal-relevant change in the bump, and **M4 does not exercise
its new path**. M4 reports `simdgroup matrix mul. = true` (Apple9), so `has_simdgroup_mm`
short-circuits to the pre-existing behaviour and OUT_PROD dispatches on the GPU exactly as
before. This run confirms the guard does not regress Apple7+; it does **not** test what the
guard was written for — the iOS simulator reporting Apple2, and A13-and-older devices, where
the intended effect is a clean decline and CPU fallback. That path needs a simulator or an
older device, which is squarely `feature/ios-build` territory.

### Verdict

The SA3 Metal gate passes. Nothing found that should block the pin bump.

PR #37, opened 2026-09-01
train: persist pre-encoded latents so --latents-dir can consume our own output

### LINKED ISSUE #35: train: persist pre-encoded latents so --latents-dir can consume our own output
`--latents-dir` is consume-only. `train_load_latent_dir` reads the `.npy`/`.json` sidecar pairs a gary4local job produced (`encoded/latents/<model>/`), which exists so a parity run can train on bit-identical data to the PyTorch reference. sa3.cpp's own pre-encode builds `lat_cache` in memory and drops it when the run ends, and there is no npy writer anywhere in the tree.

So the only way to reuse latents is to produce them with gary4local's `pre_encode.py` — Python and torch. Fine on a workstation, impossible on iOS, where every run re-encodes the whole corpus.

## Why it matters

Pre-encoding 42.6 min of audio through SAME-L on an iPhone 13 (A15) takes long enough to dominate a training session, and it happens again on every parameter change. Sweeping rank / lr / crop / DiT means paying it every time for latents that are bit-identical across all of those.

## Cache key

Latents depend on the autoencoder, not on what is usually being changed:

| changing | invalidates |
|---|---|
| DiT encoding (q4 -> f16), training base vs inference DiT | no |
| rank / alpha / lr / crop frames / steps / seed / adapter type | no |
| `variant` (SAME-S vs SAME-L) | yes |
| `autoencoder_encoding` | yes |
| `target_latent_rms` | yes |
| the audio itself | yes |

So a key of `(variant, ae_encoding, target_latent_rms, per-file audio_sha256)` covers it. `audio_sha256` is already parsed from `metadata.jsonl` and used for train/test contamination checks, so there is an existing per-file identity to hang it on.

## Shape

- write to `<dataset>/latents/<variant>-<ae_encoding>/` by default, beside the data it derives from and trivially deletable
- reuse the existing `.npy` + `.json` convention rather than a new format: it keeps gary4local parity in both directions and the reader already exists, so only a writer is missing (npy is a header plus raw f32)
- on startup, load what matches the key and encode only the misses. The pre-encode loop already does `if (lat_cache.count(stem)) continue;` — the skip logic is there, it just has nothing to populate the cache from

## Related

Pre-encode on Metal also accumulates wired memory: `encode_train_audio_to_latents` builds a fresh `ggml_gallocr` per chunk, and Metal residency sets keep buffers wired for 180 s by default, so on a 4 GB device the GPU eventually cannot allocate and resets (`kIOGPUCommandBufferCallbackErrorInnocentVictim`). `GGML_METAL_RESIDENCY_KEEP_ALIVE_S=1` works around it; hoisting the allocator out of the per-chunk call would remove the churn instead, and would apply to audio2audio, which re-encodes on every continuation. Worth considering alongside this, since a latents cache is the other half of not paying for the encode repeatedly.

betweentwomidnights:
Closes #35.

Native pre-encode now writes its latents and reuses them next run. On the 10-track ratatat set a
second run reaches step 1 in **12 s instead of 285 s**, with bit-identical losses. 27 MB of cache
for 42.6 min of audio.

Keyed on the **decoded audio**, not the source file, so it works the same whether audio comes from
the filesystem or from a host's `load_audio` hook — a file fingerprint would have served
workstations and skipped iOS, which is what the issue was filed for. Anything that invalidates
(variant, autoencoder encoding, `target_latent_rms`) gets its own directory; the key in the sidecar
is the guard. A cache that cannot be written is reported once and the run continues.

## Verified on the PC (CUDA)

Both directions, against the real gary4local/PyTorch stack rather than against our own reader:

- our `.npy` loads in numpy 2.4.6 — the interpreter `services/sa3/env` runs — with correct shape,
  dtype, order and values
- a genuine PyTorch pre-encode job (`billie-1788079527`) trains here through `--latents-dir`
- `PreEncodedDataset.__getitem__` hard-indexes only `padding_mask` and `seconds_total`; both are
  written, so our cache directories are directly trainable by the official dataset
- `n_valid` is identical to `_valid_latent_length`, not merely close

Fixed one real mismatch found this way: the reference spells the field `latent_rms_pre_norm`.

21/21 tests. `tests/latents_cache_test.cpp` covers the round-trip, the npy against the NumPy spec,
every invalidation case, and that a foreign sidecar never satisfies a cache lookup.

## Validated on Metal

- [x] M4
- [x] iPhone

## API naming, found during that validation

`config_json` took a **path** to a JSON file while both its name and its doc comment said content;
the CLI spells the same thing `--config`. It is `config_path` now. `prompt_config` was also a path
missing its suffix, next to a correctly-named `resume_path`, and maps to `prompt_config_path`
internally — renamed to match. Neither field moves or changes type, so the ABI is unchanged and an
already-compiled host keeps working; **source naming either field needs the new spelling.**

That audit also turned up two gaps: `latents_cache` / `latents_cache_dir` had never been exposed on
the C ABI at all, so the read-only-dataset case this feature exists for was only reachable through
`config_path`; and `--config` was parsed but absent from `--help`. Both fixed.

`latents_cache` follows `checkpoint_every`: 0 keeps the default, negative disables, so a
zero-initialized config still behaves.

## Validated after the rename

The iOS host was recompiled against the current header and re-checked on device; the naming
change improved things on the `sa3.cpp-ios` side too. `latents_cache_dir` and `latents_cache = -1`
are also exercised end-to-end through the C ABI by `sa3-libtrain`: a redirected cache writes to
the given directory and leaves `dataset_dir` untouched, a second run reuses it with an identical
loss, and the disable path encodes without reading or writing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -253,6 +253,10 @@ target_link_libraries(sa3-model-paths-ae-test PRIVATE sa3)
 add_executable(sa3-lora-quant-merge-test tests/lora_quant_merge_test.cpp)
 target_link_libraries(sa3-lora-quant-merge-test PRIVATE sa3)
 
+add_executable(sa3-latents-cache-test tests/latents_cache_test.cpp vendor/yyjson/yyjson.c)
+target_link_libraries(sa3-latents-cache-test PRIVATE sa3)
+target_include_directories(sa3-latents-cache-test PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/vendor/yyjson)
+
 add_executable(sa3-set-wide-row-test tests/set_wide_row_test.cpp)
 target_link_libraries(sa3-set-wide-row-test PRIVATE sa3)
 
@@ -276,6 +280,7 @@ if(BUILD_TESTING)
     add_test(NAME sa3-lora-compose-test COMMAND sa3-lora-compose-test)
     add_test(NAME sa3-model-paths-ae-test COMMAND sa3-model-paths-ae-test)
     add_test(NAME sa3-lora-quant-merge-test COMMAND sa3-lora-quant-merge-test)
+    add_test(NAME sa3-latents-cache-test COMMAND sa3-latents-cache-test)
     add_test(NAME sa3-set-wide-row-test COMMAND sa3-set-wide-row-test)
     if(Python3_Interpreter_FOUND)
         add_test(NAME sa3-model-artifacts-test
@@ -288,7 +293,7 @@ if(BUILD_TESTING)
         sa3-train-diffusion-test sa3-train-lora-test sa3-train-dit-compile-test
         sa3-train-optimizer-test sa3-train-loop-compile-test sa3-train-checkpoint-test
         sa3-train-prompt-test sa3-train-inpaint-test
-        sa3-dit-lin-functional-test
+        sa3-dit-lin-functional-test sa3-latents-cache-test
         PROPERTIES LABELS "non-gpu;training;lora")
 endif()
 
--- a/docs/TRAINING.md
+++ b/docs/TRAINING.md
@@ -27,6 +27,7 @@ sa3-train --dataset /path/to/dataset --steps 2000 \
 | `--duration SEC` | — | seconds of audio per training crop. Without it you get `--frames 512`, which is **47.6 s**. |
 | `--lora-scope full\|core` | `full` | `core` adapts only the 24 blocks' own projections (168 weights instead of 228): ~10% faster steps and ~11% smaller adapters, and in a matched 2000-step run the loss curve was within 0.0013 of `full` throughout. |
 | `--t5-encoding ENC` | auto | text-encoder precision, resolved apart from `--encoding`. Auto prefers `F16`, which is equivalent to `F32` at half the size (1074 → 537 MiB) and takes small-music's training peak from 2.28 to 1.78 GiB. `q8_0` saves another 0.26 GiB for a small, audible-on-paper cost. **Pick it for memory, never for speed:** encoding the caption is `t5_cond` in `SA3_TRAIN_PROFILE=1`, and on CUDA/medium that is **9 ms of a 1070 ms step — 0.84%** — so `q8_0` moves the step by −1 ms, within noise of `f16` (see below). |
+| `--latents-cache BOOL` | `true` | keep the pre-encode output and reuse it next run. Latents depend on the autoencoder and `--target-latent-rms` and on nothing the DiT or the adapter does, so a sweep over rank / lr / frames / steps / seed hits the cache every time. On the 10-track ratatat set a second run starts in **12 s instead of 285 s**, with bit-identical losses. Written to `<dataset>/latents/<variant>-<ae encoding>[-rms<target>]/`, or `--latents-cache-dir DIR` for a host whose dataset directory is read-only. A cache that cannot be written is reported once and the run continues. |
 | `--evict-text-encoder BOOL` | `false` | drop the text encoder between batches of captions instead of holding it for the whole run. It encodes a window of upcoming captions in one pass (63 of them, 47.6 MiB) and frees T5 until that window runs out, giving back **285 MiB at `q8_0`** for one encoder reload per window. The same reasoning as `--t5-encoding` taken to its conclusion: T5 is under 2% of a step but 15-19% of resident memory, so on a memory-bound host it is the first thing that should leave. **Quantizing past `q8_0` is not the alternative** — `q4_k_m` saves 79 MiB and was measured to cost audible quality. Needs pre-encoded latents (the default) and is silently inert without them. Off by default: a desktop has no reason to pay the reload. |
 | `--encoding ENC` | `f16` | base precision. Quantized bases (`q4_k_m`, `q8_0`, …) train on every backend and are both smaller and faster — on an M4 a `q4_k_m` medium base ran 2.77 s/step against 3.28 s/step for f16, on 962 MiB instead of 2773 MiB. |
 
@@ -37,6 +38,23 @@ Two things that are easy to miss:
   both are given. Shorter crops train faster and use less memory; the reference regime is ~47 s.
 - **MP3 datasets need `ffmpeg` on `PATH`.** Decoding shells out to it (see [Dataset](#dataset)).
   WAV already at 44.1 kHz does not — it is read natively.
+
+### Reusing latents
+
+The cache is written in the same `.npy` + `.json` convention `--latents-dir` reads, so it goes
+both ways: a directory this wrote can be handed to `--latents-dir`, and a pre-encode from
+gary4local, underfit, or the official PyTorch trainer can be trained on directly. The arrays are
+`[latent, frames]` C-order float32 with a `padding_mask` in the sidecar.
+
+The two flags are not interchangeable, and the difference is the point:
+
+- `--latents-dir` is **explicit**: it trains on exactly what you point it at, no questions asked.
+  That is what a parity run against a PyTorch reference needs.
+- `--latents-cache` is **checked**: an entry is only reused when the autoencoder, the loudness
+  target, and a fingerprint of the decoded audio all still agree. Anything else re-encodes.
+
+So a foreign pre-encode never satisfies a cache lookup by accident — it has no key in its sidecar
+and is skipped. Feeding one to the trainer is something you ask for by name.
 
 Training prints what it is actually doing at startup, which is worth a glance before walking away:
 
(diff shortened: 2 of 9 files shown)


betweentwomidnights:
## Validated on the M4 (Metal)

Ticks the **Metal / M4** box. Both directions, against the installed gary4local rather than
against our own reader: `numpy 2.4.6`, `mlx.core 0.31.2`, `torch 2.7.1` in
`venvs/sa3` — the same numpy the PC checked, but a genuinely different encoder stack.

**Our `.npy` → gary4local.** All 10 files pass 114 checks: `float32`, C-order, 2-D, finite,
`len(padding_mask) == T`, data on a 64-byte boundary. `np.load(..., mmap_mode="r")` returns an
array identical to a plain load, which matters because that is exactly how
`scripts/train_mlx_lora.py` reads latents — the MLX trainer's read path is covered, not just
numpy's.

**gary4local → us.** `--latents-dir` against a real job (`ratatat-3-1784005242`) loaded all 10
files and trained, taking prompts from their sidecars.

**Cache reuse.** Second run: `10 reused, 0 written`, 22 s wall. Losses are bit-identical to every
printed digit — `loss=0.835824 gnorm=0.1549` and `loss=0.707654 gnorm=0.1006` in both runs.

**`--evict-text-encoder`** works here too (`window 63 caption(s), 47.6 MiB`), losses unchanged
against the non-evicting run.

**42/42 ctest**, including `sa3-latents-cache-test`.

### The padding convention differs, compatibly

Worth writing down since it looks alarming at first: for the same track gary4local stores
`[256, 6459]` and we store `[256, 3233]`. That is not a latent-rate mismatch. Their
`padding_mask` sums to exactly 3233, their valid region correlates with our array at **0.99988**,
and valid-region RMS agrees to five decimals (0.925763 vs 0.925767). They pad to a fixed length
and mark validity in the mask; we write only valid frames with an all-ones mask. Their masked-out
tail is silence-latent padding, not zeros. Both are legal because `padding_mask` is authoritative
on both sides.

### Recommend adding `prompt` to our sidecar

Inheritance is currently **asymmetric**. Their sidecars carry `prompt`, so `--latents-dir` on a
gary4local job trains with captions and no extra files. Ours does not, so a gary4local/underfit
job pointed at one of our cache directories would get latents and no captions. Aligning on the
`prompt` field that stable-audio-3, underfit and gary4local already use would close that.

Also in theirs and not ours, for whoever picks this up to rule in or out:
`path`, `relpath`, `src_relpath`, `seconds_start`, `audio_samples`, `latent_shape`.

### Pre-encode cost on Metal

Re-measured properly. Two earlier passes at this were wrong and are superseded by the numbers
below: the first compared our loudness-fix run against a gary4local run with the fix *off* (so
the `audio_gain_applied = 35/64` "they don't iterate" inference should be ignored), and both were
taken while gary4local was running concurrently, which inflated them by 1.3-1.6x.

Method: one track (Montanita, 300.3 s), interleaved `A/B/A/B` with 90 s cooldowns, fresh cache
each run, per-line timestamped logs so the figure is `AE load + encode` between the `lora scope`
and `[pre-encode]` lines rather than wall-clock. Within-config spread came out at 1.6% and 2.9%,
so this is comfortably clear of this machine's thermal drift.

| | AE load + encode | note |
|---|---|---|
| ours, `--target-latent-rms 0.9` | **385.5 s** | 380 / 391, 4 rounds |
| ours, no RMS target | **62.5 s** | 62 / 63, 0 rounds |
| gary4local, loudness fix on | 187.52 s | `encode_elapsed_seconds` |
| gary4local, loudness fix off | 33.18 s | `encode_elapsed_seconds` |

Two conclusions, and the first retracts what I said earlier:

**Both implementations iterate for the loudness target, at a similar cost** — ours 6.2x, theirs
5.7x. There is no anomaly in how we converge, and nothing to fix there.

**The real gap is a flat ~2x on the encode itself**, and it is consistent across both configs:
1.88x without the loudness fix, 2.06x with it. Ours includes AE load where theirs is encode-only,
so the true ratio is a little under 2x. That is a much smaller gap than the 3x I quoted from the
contaminated run.

Nothing here blocks the PR — the cache means it is paid once. It is worth knowing that a cold
cache with `--target-latent-rms` on Apple silicon is ~6.5 min/track, so tens of minutes for a
10-track set, and that this is inherent to the loudness fix rather than a Metal pathology.

### Note on `PreEncodedDataset`

I drove our cache directory through gary4local's own `PreEncodedDataset` (10 files, cropped to
`(256, 512)`, mask and `seconds_total` honored) — **but** `dill` is not installed in any of the six
gary4local venvs, so `stable_audio_3.data.dataset` cannot import in the shipped app and I stubbed
it to run the check. Read that result as "the format contract is satisfied", not "the app does
this today". The path that is actually live on macOS is the MLX trainer's own `np.load`, which is
validated above. The missing `dill` may be worth a look on the gary4local side independently.

### Not covered

The MLX trainer end-to-end on our cache (needs a job config and checkpoint); iPhone.

betweentwomidnights:
## Validated on iPhone 13 (A15, Metal)

Both remaining boxes. Branch at `f4fd55a`, ggml `19c5421c`, small-music q4 DiT / F16 autoencoder, 10-track ratatat set (42.6 min).

**Cache write + reuse.** First run pre-encodes the corpus in ~5 min (2:59 -> 3:04); a second run with identical settings reaches step 1 in a few seconds. Proportionally a much larger saving than the 285s -> 12s measured on CUDA, which is expected -- SAME-S pre-encode is far slower here, and this is the machine the issue was filed for.

**Read-only / write-failure path.** Pointing `--latents-cache-dir` at a path outside the sandbox gives `not caching (cannot create /latents-should-fail: Operation not permitted)`, reported once, and pre-encode proceeds normally. Degrades exactly as documented.

**Cache layout.** `<dataset>/latents/small-music-F16/`, keyed on variant + autoencoder encoding as intended. 27 MB for the corpus.

### npy verified independently

Parsed with a from-scratch reader rather than the one that wrote it:

    magic     b'\x93NUMPY'  version 1.0
    header    {'descr': '<f4', 'fortran_order': False, 'shape': (256, 3233)}
    data      3,310,592 bytes = 256 * 3233 * 4   exact
    padding   header ends 64-byte aligned        per spec

`shape[1] = 3233` is written at `n_valid`, not padded to the 600 s `kPreEncodePadSeconds` grid: the track is 300.28 s and `ceil(300.28 * 44100 / 4096) = 3233`. That is why the corpus fits in 27 MB.

Sidecar carries `padding_mask` (3233 entries, last = 1), `seconds_total` = 300.28, `latent_rms_pre_norm`, `source_fingerprint`, `autoencoder`, `target_latent_rms`, `cache_version`. `n_valid` is absent and correctly so -- `train_load_latent_dir` derives it from `padding_mask`.

### One host-facing snag, worth solving before this merges

Reaching `--latents-cache-dir` from a library host cost a failed run: `config_json` is a **path**, but reads as JSON content.

    const char* dataset_dir;     /* REQUIRED */
    const char* output_dir;
    const char* latents_dir;
    const char* resume_path;
    const char* config_json;     <- the only path-valued field not named _dir or _path

Every other path in `sa3_train_config` is signposted by its name; this one is not, and the comment ("optional JSON of any train config key") points the wrong way. All of them are `const char*`, so nothing catches it -- passing the JSON body yields `cannot open config file`.

`config_json` predates this PR, but this PR makes it load-bearing: `latents-cache` and `latents-cache-dir` have no struct fields, so `config_json` -- and therefore writing a temp file -- is the only route for a library host. A sandboxed host is exactly who needs `--latents-cache-dir`, since the dataset directory may not be writable.

Two options, not exclusive: rename to `config_json_path` (restores the struct's own convention rather than relying on the comment being read), and/or promote the two cache options to real fields the way `evict_text_encoder` was.

betweentwomidnights:
Re-validated on the iPhone 13 against `16e005c`. The renames and the two new fields are good from a host's side.

Reaching the cache used to mean knowing `config_json` was a path, building JSON, writing a temp file, passing its path and cleaning up -- nine lines, with a failure mode (temp write) unrelated to training. It is now:

    cfg.latents_cache = latentsCache ? 0 : -1

plus a `withCString` for `latents_cache_dir`. The `checkpoint_every` convention is the right call for a C ABI where hosts memset the struct: a bool would have made "off" the zero value and silently disabled the cache for every existing host.

The confirmation run **reused latents written before the ABI change**, so the rename did not disturb the cache key or the sidecar format -- no invalidation from the API work, which is worth knowing given the cache is keyed on content rather than on anything the API touches.

`prompt_config_path` was the better catch. It was already spelled that way in `TrainConfig`; the C ABI was the only place it disagreed.

Confirming the earlier note in practice: pulling the branch alone is not enough for a host, since `vendor/include/libsa3.h` is staged by the build script. The build failed on `latents_cache` (genuinely new) rather than on the renames -- consistent with the commit's claim that the ABI is unchanged and only source spelling moves.

PR #43, opened 2026-09-14
Establish libsa3 C ABI V1

betweentwomidnights:
## Goal

Establish the production-candidate libsa3 C ABI V1 before sa3.cpp has a public release and before more frontends depend on the historical entry points.

## ABI shape

- one `sa3_get_api(SA3_ABI_VERSION_1)` symbol and a size-tagged inference function table
- a separately versioned `sa3_get_training_api(SA3_TRAINING_ABI_VERSION_1)` table
- size-tagged C11 request, result, error, adapter, audio, loudness, continuation, training config/callback/result structs
- frozen initial compatibility prefixes with append-only growth rules
- typed statuses, initialized defaults, and stable error ownership
- library-owned Generate / Transform / Continue duration planning, SAME-S alignment, continuation headroom, splicing, resampling, and exact trimming
- planar/interleaved input and library-owned planar output
- adapter import/conversion and strided per-request adapter arrays
- training progress/log callbacks, cancellation, valid partial results, and an optional owned host-audio bridge
- historical entry points retained as non-contract compatibility shims for the initial V1 release

The distribution-shift parameter array uses all zeroes as “defaults for the selected shift.” This lets a client switch LogSNR / Flux / Full / None without duplicating model tuning constants and prevents one shift's initialized parameters from leaking into another.

See `docs/C_ABI_V1.md` for the contract and migration rules.

## Integration gate

Do not merge until this exact branch has final sign-off from all controlled consumers:

- [x] `sa3.cpp-iplug2-demo`: VST3 and REAPER source migration plus Windows Generate / Transform / Continue, model, creative LoRA, decoder LoRA, exact-duration, and timeline tests
- [x] `sa3.cpp-iplug2-demo`: final CUDA and Vulkan VST3/REAPER rebuilds against `a690e1c`; bundled runtime hashes matched and both VST3 bundles passed 47/47 validator tests
- [x] `sa3.cpp-iplug2-demo`: final installed cancellation/frugal-lifecycle smoke test; REAPER cancellation completed cleanly and resident-mode shutdown released the host/GPU allocation
- [x] `sa3-ableton-extension`: embedded source migration plus Windows Generate / Transform / Continue, model, creative LoRA, decoder LoRA, exact-duration, and timeline tests
- [x] `sa3-ableton-extension`: CUDA and Vulkan packages against `a690e1c`; CLI/native Full-shift zero-tail request verified byte-identical
- [x] `sa3-ableton-extension`: final cancellation and resident/frugal lifecycle smoke test
- [x] `sa3.cpp-ios`: inference and Training V1 source migration on `feature/c-abi-v1` (`fb933e5`)
- [x] `sa3.cpp-ios`: validated against `a690e1c` with device/simulator builds, V1 contract and 20 non-GPU tests, exact-duration inference, adapter conversion/stride, cancellation, audio ownership, final-adapter registration, and a completed training run
- [x] legacy ABI decision: retain the historical entry points as non-contract compatibility shims for the initial V1 release; the versioned API tables are the production contract

## Verification

- [x] CUDA and Vulkan Release shared-library builds on Windows
- [x] pure-C inference + training ABI contract builds on CUDA and Vulkan
- [x] adapter-stride walk coverage through the public API
- [x] future-caller tail preservation checks
- [x] all 34 configured CUDA-tree CTests
- [x] packaged Ableton DLL hashes match the tested core builds


betweentwomidnights:
Desktop integration branches are now live and revalidated against core commit 2e0871e:

- iPlug2 VST3 + REAPER: https://github.com/betweentwomidnights/sa3.cpp-iplug2-demo/pull/4
- embedded Ableton backend: https://github.com/betweentwomidnights/sa3-ableton-extension/pull/1

The MacBook/iOS consumer remains the intentional final gate, including adoption of the new Training V1 table.

betweentwomidnights:
Revision alignment update: core ABI V1 is now at `82ca416`. Training has explicit host-audio ownership/release semantics and its own shared-error initializer. Frozen minimum prefixes, future-caller tail checks, and a real padded/strided adapter-array contract test are in place. All 24 Windows CTests pass. Both desktop consumers were rebuilt against this exact core: iPlug2/REAPER `c4b93ae`, embedded Ableton `b1092f4`. The PR remains draft pending runtime regression checks and final Mac/iOS inference + training sign-off.

betweentwomidnights:
Two notes out of the `sa3.cpp-ios` migration. Raising them now rather than with sign-off, since both look like things to check while updating the Ableton extension and the iPlug2 demo.

**Coverage gap in the ABI's own suite.** `tests/libsa3_v1_contract.c` asserts `request.adapter_stride == sizeof(sa3_adapter_v1)` but never passes an actual adapter array, so the strided walk itself has no coverage. Worth a case there given all three frontends pass adapters.

**The operation + `duration_seconds` model removes real client code.** On iOS it deleted more than it added: `continueTake()` no longer places an inpaint window, `create()` no longer converts seconds to frames, and the create panel's latent-frame readout is gone entirely — V1 rounds up and to even, so 11.9 s is 130 frames where the app had been reporting 128. Any frontend still doing its own frame math, window placement, or trimming is carrying a second copy of arithmetic V1 now owns, and that copy can drift. Worth grepping both of the other frontends for it during their migration.

Holding the iOS validation comment until the branch settles — happy to re-run it here against whatever this lands on.

betweentwomidnights:
Addressed both iOS migration notes in `82ca416` (with the corresponding iPlug cleanup in `c4b93ae`). The C contract now sends a two-entry padded adapter array through the public `generate` function with `adapter_stride > sizeof(sa3_adapter_v1)`. A tightly-packed decoy and a distinct invalid field in the real second entry prove the walker uses the declared stride. Adapter validation also now checks multiplication overflow, entry size versus stride, empty names, and non-finite strengths with indexed diagnostics.

The desktop arithmetic audit found no duplicate frame, SAME-S alignment, inpaint-window, or output-trim math in either C ABI path. iPlug/REAPER pass operation + seconds directly; the VST only translates its total-length UI into Continue seconds-to-add, while REAPER retains timeline placement/maximum-duration checks. Embedded Ableton likewise passes operation + seconds directly. Its remaining `target_samples`/inpaint arithmetic belongs only to the separate legacy HTTP backend contract. One dead VST output-trim parameter was removed.

Also added `error_init` to the Training V1 table so training-only hosts do not depend on the inference table. All 24 Windows CTests pass, both iPlug targets rebuild, and the embedded Ableton package rebuilds with matching core DLL hashes. The iOS branch should now restage core `82ca416`, use `training.pointee.error_init`, validate both table minimum sizes, and rerun simulator/device checks before sign-off.

betweentwomidnights:
**`sa3.cpp-ios` validated against `a690e1c`.**

Migration is `betweentwomidnights/sa3.cpp-ios@fb933e5` on `feature/c-abi-v1`, held off that repo's `main` until this merges.

**On device** (iPhone, small-music):
- context lifecycle — create / unload / destroy
- Generate with a trained DiT adapter above 0 strength, f16 across all three nets
- cancellation mid-generate
- Training V1 end to end, including final-adapter registration, re-run after `82ca416` changed the training table's shape

**On the MacBook**, through the V1 table in C:
- exact durations — Generate 4.0s → 176400 samples; Continue 4.0s + 3.0s → 308700; Transform ignores `duration_seconds` and follows its input
- continuation splice fires on Continue and reports through `sa3_result_v1`
- `convert_lora` with a null JSON sidecar
- strided adapter array, routed to the decoder by gguf metadata
- `should_cancel` → `SA3_STATUS_CANCELLED_V1`
- `result_free`, including the documented double-free safety
- `sa3-lib-v1-contract` plus 20/20 non-GPU CTests

**Simulator is build-and-link only** — it has not been a useful runtime target for this app even on an M4, so no runtime claim is made. It did not link at all on an Intel host until the iOS project's library search path was made arch-aware (`vendor/lib-sim` is arm64; x86_64 needs `lib-simx64`). That fix is local to the iOS repo.

**On `a690e1c` specifically:** confirmed each shift now resolves its own defaults from all-zero params — LogSNR, Flux, Full and None produce four distinct schedules, and forcing explicit params produces a fifth. LogSNR's resolved schedule is identical to what it produced before the fix, so the default path is unchanged. This repo's picker offered Flux/Full/None all along and was silently running LogSNR's parameters for them.

One observation for after the merge, not a blocker: `82ca416` inserted `error_init` into the training table ahead of `config_init` rather than appending it. Correct pre-merge and it is what the draft is for, but it is the last free window to reshape those tables — once the three frontends ship against V1, the contract's append-only rule starts binding for real.

PR #44, opened 2026-09-15
libsa3: retire the legacy ABI, and stop routing V1 through it

betweentwomidnights:
## Summary

This removes the retired, unversioned libsa3 ABI and makes the published V1 tables the only C embedding surface.

- Inference now maps `sa3_request_v1` directly to pipeline parameters instead of routing through legacy request structs.
- Training and LoRA conversion use only their V1 table entry points.
- The library exports `sa3_get_api` and `sa3_get_training_api`.
- The pure-C tools and embedding documentation now use V1 throughout.

## Compatibility

This does not invalidate the Ableton v0.2.0 or iPlug2 v0.4.0 releases already shipped: each release bundles the libsa3 binary it was built with. Their consumers resolve `sa3_get_api`, not any removed legacy symbol.

For future rebuilds, both published V1 headers are unchanged from `main`, and neither V1 function table changed shape or ordering.

## Behavioral safeguards

The refactor keeps the complete generation operation behind one C ABI exception boundary, including request translation and allocations. It also preserves the V1 behavior frontends can observe: unresolved model sets return `SA3_STATUS_MODEL_ERROR_V1`, and explicitly empty configuration strings are not silently replaced with defaults.

`sa3-lib-v1-contract` pins those behaviors. `sa3-lib-v1-baseline` provides a model-backed fingerprint over Generate, Transform, Continue, loudness, residency, all distribution shifts, cancellation, invalid requests, and an optional adapter. Runtime version provenance is written to stderr so intentional version bumps do not contaminate the behavioral diff.

## Verification

- Full Windows CPU build completed.
- CTest: 24/24 passed.
- CPU baseline output matched `main` for all 11 core scenarios; the only pre-normalization difference was the intentional libsa3 runtime version change from 6 to 7.
- The optional adapter scenario was not run because no compatible adapter GGUF was available in this checkout.
- `sa3.cpp-iplug2-demo` was rebuilt against this branch: standalone app, VST3, and REAPER extension all completed successfully.
- All three rebuilt consumers bundled the branch DLL byte-for-byte and contain `sa3_get_api` without removed legacy symbol names.
- The standalone app remained alive through a startup smoke check and was then closed deliberately.

The iOS consumer was not rebuilt as part of this PC-side pass.


betweentwomidnights:
Both findings confirmed and fixed in `028086c`. Thanks — the first one is a host crash, not a style point, and I had it wrong.

**Exception boundary.** Correct diagnosis. The old code wrapped everything in `sa3_generate_impl`'s single try, and the V1 layer above it only marshalled into POD C structs, so nothing allocated outside the guard. My rewrite moved the allocating work — `p.prompt`, `p.negative_prompt`, `p.dist_shift`, two `std::function` assignments — above the first `catch`.

Fixed the way you recommended rather than by patching the holes: `v1_generate` is now nothing but the boundary, and the body and the run are internal and deliberately unwrapped. Scattered local catches are exactly what let the gap open, so there is now one place to look.

The allocation-in-handler point was right too, and slightly worse than stated: `set_v1_error` took a `const std::string&`, so *every* error path allocated, including the out-of-memory one. There is now a literal path that writes straight into the caller's fixed buffer, the cancellation check uses `strstr` instead of a `std::string` search, and `context_create` / `convert_lora` / `training_run` handlers use it too. `convert_lora` also gained a `bad_alloc` handler it never had.

**Observable V1 changes.** Both real, both restored as published rather than accepted:

- unresolvable model set returns `MODEL_ERROR_V1` again, not `IO_ERROR_V1`
- empty `variant` / `dit_encoding` / `adapters_dir` are passed through again rather than selecting defaults. NULL still falls back. `models_dir` keeps treating empty as unset, which is what it always did — that one was not a change.
- `context_create` passes `config->device` through as the raw pointer, so an empty device string reaches the backend exactly as before

You were right that these might be better semantics. They are not better enough to be worth changing silently underneath three shipped frontends, and "unset" versus "explicitly empty" is a real distinction a host may be relying on.

They are now pinned in `libsa3_v1_contract.c` — a nonexistent models dir must be `MODEL_ERROR_V1`, an empty variant must fail rather than quietly become `medium`, and negative `cpu_threads` must be `INVALID_ARGUMENT_V1`. No models needed, so it runs in CI. The next refactor has to mean it.

**Re-verified:** equivalence against `9acbc8f` over the same 12 scenarios is still bit-identical, and 20/20 non-GPU CTests pass with the three new cases.

One thing I could not test directly and would value your view on: I have no way to force `bad_alloc` at those sites, so the boundary is verified by construction and inspection rather than by a fault-injection test. If the plugin host side has a way to exercise that, it is worth doing once.

betweentwomidnights:
Neither, and that was my mistake — it was never in a repo at all.

The 12-scenario harness lived only in a scratch directory on the MacBook. Saying "happy to attach it if useful" in the PR description was the wrong call: a harness that is the entire evidence for "this refactor changed nothing" has to be reviewable, and a temp directory on one machine is neither reviewable nor durable.

It is now `tools/sa3-lib-v1-baseline.c` in `a0073a1`, on this branch.

It covers the half `sa3-lib-v1-contract` cannot. The contract test needs no models and checks the ABI's shape. This one runs real generations and prints what came back — exact sample counts, every reported metadata field, and an FNV-1a hash of the decoded audio — across all three operations, non-default loudness, frugal residency, every distribution shift on its resolved defaults, explicit shift parameters, an adapter array, cancellation, and two invalid requests. Capture before, capture after, diff.

```sh
export SA3_MODELS_DIR=/path/to/models
export SA3_BASELINE_DEVICE=cpu SA3_BASELINE_SECONDS=1.0 SA3_BASELINE_STEPS=2
sa3-lib-v1-baseline > before.txt      # then make the change and rebuild
sa3-lib-v1-baseline > after.txt
diff before.txt after.txt
```

It needs a model set, so it is a tool rather than a CTest. About 100s for the whole set at those settings. The optional argument is an adapter gguf for the medium/SAME-L family; omit it to skip that one scenario.

**Run it on CPU, and confirm that on your own hardware before trusting a diff from it.** Metal on this AMD Radeon Pro 5300M is not run-to-run reproducible — the same binary and the same seed hash differently every run. That cost me a wrong conclusion: my first comparison showed audio differing everywhere and looked exactly like a broken refactor, until the control (the identical binary, run twice) also disagreed with itself. If your GPU path turns out to be reproducible, a GPU diff is fine; I would just verify it with two unchanged runs first rather than assume either way.

Verified the committed tool reproduces the pre-refactor fingerprint exactly, so what is on the branch is the same measurement the earlier claims rested on, not a re-implementation of it.

`docs/C_ABI_V1.md` now describes both tools and which half each one covers.

